PR #30, opened 2023-08-01
Fix destroy with stuff from other apps

gmrabian:
### Description
- Remove `yarn install` step as we don't need that right now
- Fix bucket script to empty buckets with a large amount of items
- Make stage disassociation conditional on existence
- Remove hanging log groups

### How to test
Create a branch
Run destroy
Weeeeeeeeee

### DIFF
--- a/scripts/destroy.sh
+++ b/scripts/destroy.sh
@@ -1,19 +1,6 @@
 #!/bin/bash
 set -e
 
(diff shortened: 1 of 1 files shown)


karla-vm (approved):
LGTM!!

karla-vm (approved):

PR #98, opened 2023-09-14
MDCT-2794: Update submission name

trancjessica:
### Description
This change updates the name of the Work Plan as displayed on the Dashboard page. (i.e. revises it from `<program name> // <submission name>` to `{State} <program name> {<year> - Period <period>})`

Note: The requirement is to add the "State" to the submission name. I pulled the state from the `report`. Given that the mocks show that {State} should also appear in the **Page Title**, I assumed this was the correct value to fetch? Open to thoughts!! 🙏

Screenshot 2023-09-14 at 11 40 20 AM

### Related ticket(s)

MDCT-2794

---
### How to test

1. Log in as a state user
2. Navigate to work plan page (i.e. /wp)
3. Create a new work plan 
4. Confirm that the work plan name matches the requirements ({State} <program name> {<year> - Period <period>})

### Important updates

---
### Author checklist

- [✅ ] I have performed a self-review of my code
---

### DIFF
--- a/services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
+++ b/services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
@@ -8,9 +8,12 @@ import {
   ReportType,
   TableContentShape,
 } from "types";
-import { convertDateUtcToEt } from "utils";
+import { convertDateUtcToEt, calculatePeriod } from "utils";
 // assets
 import editIcon from "assets/icons/icon_edit_square_gray.png";
+
+// Year will be displayed as part of the report name
+const year = new Date().getFullYear();
 
 export const DashboardTable = ({
   reportsByState,
@@ -43,7 +46,10 @@ export const DashboardTable = ({
         )}
         {/* Report Name */}
         <Td sx={sxOverride.programNameText}>
-          {report.programName ?? report.submissionName}
+          {report.state} {report.submissionName ?? report.programName} {"{"}
+          {year} {" - Period"}{" "}
+          {calculatePeriod(convertDateUtcToEt(report.createdAt))}
+          {"}"}
         </Td>
         {/* Date Fields */}
         <DateFields report={report} reportType={reportType} />
--- a/services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
+++ b/services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
@@ -45,12 +45,7 @@ export const DashboardTable = ({
           <Td></Td>
         )}
         {/* Report Name */}
-        <Td sx={sxOverride.programNameText}>
-          {report.state} {report.submissionName ?? report.programName} {"{"}
-          {year} {" - Period"}{" "}
-          {calculatePeriod(convertDateUtcToEt(report.createdAt))}
-          {"}"}
-        </Td>
+        <Td sx={sxOverride.programNameText}>{getReportName(report)}</Td>
         {/* Date Fields */}
         <DateFields report={report} reportType={reportType} />
         {/* Last Altered By */}
@@ -130,6 +125,14 @@ interface DashboardTableProps {
   releasing?: boolean | undefined;
   sxOverride: AnyObject;
 }
+
+const getReportName = (report: ReportMetadataShape) => {
+  const reportName = report.submissionName
+    ? report.submissionName
+    : report.programName;
+  const period = calculatePeriod(convertDateUtcToEt(report.createdAt));
+  return `${report.state} ${reportName} {${year} - Period ${period}}`;
+};
 
 export const getStatus = (
   reportType: ReportType,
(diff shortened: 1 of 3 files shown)


services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
```diff
@@ -43,7 +46,10 @@ export const DashboardTable = ({
         )}
         {/* Report Name */}
         <Td sx={sxOverride.programNameText}>
-          {report.programName ?? report.submissionName}
+          {report.state} {report.submissionName ?? report.programName} {"{"}
+          {year} {" - Period"}{" "}
+          {calculatePeriod(convertDateUtcToEt(report.createdAt))}
+          {"}"}
```
gmrabian, line 52: since this is a little complex you could consider moving this out to a function within this file and just calling it here like
```
{getReportName(report)}
```
or something. Then you can also generate the year variable within that function. It doesn't really matter in the end, just makes reading this component a little cleaner!

services/ui-src/src/utils/other/time.ts
```diff
@@ -14,6 +14,24 @@ export const noon: TimeShape = {
...
+};
+export const endPeriodOne = (date: string) => {
+  return new Date(new Date(date).getFullYear(), 5, 30);
+};
+export const beginningPeriodTwo = (date: string) => {
+  return new Date(new Date(date).getFullYear(), 6, 1);
+};
+export const endPeriodTwo = (date: string) => {
+  return new Date(new Date(date).getFullYear(), 11, 31);
+};
```
gmrabian, line 33: periods in comment don't match functions (5/30 vs 6/30 end date for period 1)

services/ui-src/src/utils/other/time.ts
```diff
@@ -141,6 +159,16 @@ export const checkDateRangeStatus = (
...
 };
 
+export const calculatePeriod = (dueDate: string) => {
+  const dueDateAsDate = new Date(dueDate);
+  if (
+    dueDateAsDate >= beginningPeriodOne(dueDate) &&
+    dueDateAsDate <= endPeriodOne(dueDate)
+  )
+    return 1;
+  return 2;
```
gmrabian, line 169: are these used for anything other than being part of the report name? If not I'd consider returning them as a string rather than an integer, just so that we don't have any confusion with these values being used for calculations and/or 1=true, 2=? kind of thing

gmrabian (commented):
Left some comments to clear things up!

services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
trancjessica: good point! I will create a function for this! 😄

services/ui-src/src/utils/other/time.ts
trancjessica: ooh nice catch, will fix!

services/ui-src/src/utils/other/time.ts
trancjessica: i don't believe they are used for anything else, but I will double check with product. I think making them strings is a good idea, regardless! will revise :)

services/ui-src/src/utils/other/time.ts
trancjessica: upon looking at this further, i think my comments do match the function! I can revise the comments to make it more clear. Dates are zero-indexed, so new Date(2023, 5, 30) represents June 30th, 2023

services/ui-src/src/utils/other/time.ts
```diff
@@ -14,6 +14,24 @@ export const noon: TimeShape = {
...
+export const endPeriodOne = (date: string) => {
+  return new Date(new Date(date).getFullYear(), 5, 30);
+};
+export const beginningPeriodTwo = (date: string) => {
+  return new Date(new Date(date).getFullYear(), 6, 1);
+};
+export const endPeriodTwo = (date: string) => {
+  return new Date(new Date(date).getFullYear(), 11, 31);
+};
+
```
ntsummers1, line 34: I bet you could actually reduce this all down to something like 
```suggestion
return Math.ceil((new Date(date).getMonth() + 1) / 6);
```
That would return the period of either 1 or 2 based on the current month.
The way you've written this though is definitely a lot more readable though so its totally up to you!

ntsummers1 (approved):

services/ui-src/src/utils/other/time.ts
trancjessica: This is really simple! Thanks for the suggestion! I will implement this. I think the comments are clear enough to understand what the period date ranges are

ntsummers1 (approved):
LGTM!

keeysnc (approved):

cassandradanger (commented):
one small thing - you might want to check with design on this - usually the curly brackets indicate dynamic data, I'm guessing they won't want them showing in the end result

PR #123, opened 2023-09-26
[MDCT-2958]: Have repeatedFields created with form

ntsummers1:
### Related ticket(s)

MDCT-2958

---
### How to test

Create a report
Go to Page 2 of State and Territory Initiatives or page 3
Use the modal and click yes to see 12 fields that are repeated from 1 wp.json field!
Check to see if they save
Check to see if they hydrate when opening the modal
Check to see if they clear if you swap to No in the choice list
Check to see if they populate the entitycard correctly

---
### Author checklist

- [x] I have added [thorough](https://bit.ly/3zPrxuZ) tests, if necessary

### DIFF
--- a/services/ui-src/src/components/cards/EntityCardBottomSection.tsx
+++ b/services/ui-src/src/components/cards/EntityCardBottomSection.tsx
@@ -1,12 +1,10 @@
-// components
-import { Text } from "@chakra-ui/react";
 // utils
 import { AnyObject } from "types";
 
 export const EntityCardBottomSection = ({ entityType }: Props) => {
   switch (entityType) {
     default:
-      return <Text>{entityType}</Text>;
+      return <></>;
   }
 };
 
--- a/services/ui-src/src/utils/reports/entities.ts
+++ b/services/ui-src/src/utils/reports/entities.ts
@@ -1,9 +1,12 @@
 import { EntityShape, OverlayModalEntityTypes, AnyObject } from "types";
 
 const getRadioValue = (entity: EntityShape | undefined, label: string) => {
-  return entity?.[label].value;
+  return entity?.[label]?.[0].value;
 };
 
+const getQuarterValues = (entity: EntityShape | undefined, value: string) => {
+  return entity?.[value];
+};
 export const getFormattedEntityData = (
   entityType: string,
   entity?: EntityShape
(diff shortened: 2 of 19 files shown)


keeysnc:
@ntsummers1 For page 2, if you select yes and fill out the quarters should they be displayed on the cards?

keeysnc (dismissed):
Tested on local. Nice work!

services/ui-src/src/components/cards/EntityCardBottomSection.tsx
```diff
@@ -1,12 +1,10 @@
-// components
-import { Text } from "@chakra-ui/react";
 // utils
 import { AnyObject } from "types";
 
 export const EntityCardBottomSection = ({ entityType }: Props) => {
   switch (entityType) {
     default:
-      return <Text>{entityType}</Text>;
+      return <></>;
```
gmrabian, line 7: what is this file doing if it's default is to return a fragment?

gmrabian (changes requested):
Found these things and was curious about them. Let me know if these are covered elsewhere. thanks!

Do we want Quarters to show up if they choose 'No'?

Screenshot 2023-09-27 at 2 22 46 PM

Card marks itself incomplete if I choose "Other"

https://github.com/Enterprise-CMCS/macpro-mdct-mfp/assets/57802560/a6aab33c-85eb-4243-9385-b577b53f2ce0

services/ui-src/src/components/cards/EntityCardBottomSection.tsx
ntsummers1: This will be for PDF work that we haven't started or brought over from MCR. It's simply living there for now so we can add to it in the future.

britt-mo:
@gmrabian Nick fixed the quarters showing up when the answer is 'no' in this PR, the `other text` bug is getting fixed in a different PR.

gmrabian (approved):
Looks good locally! Thanks for responding to my questions 😄

services/ui-src/src/utils/reports/entities.ts
```diff
@@ -1,9 +1,27 @@
 import { EntityShape, OverlayModalEntityTypes, AnyObject } from "types";
 
 const getRadioValue = (entity: EntityShape | undefined, label: string) => {
-  return entity?.[label].value;
+  return entity?.[label]?.[0].value !== "Other, specify"
```
trancjessica, line 4: [nit, non-blocking]: would it be worthwhile to define a variable for `entity?.[label]?.[0].value` such as 
`const radioValue = entity?.[label]?.[0].value`? It might make it easier to read this code? 🤔

britt-mo: yeah, I could see `radioLabelValue` being a good name to use. This function is largely lifted from MCR so we didn't do much to change it.

PR #143, opened 2023-10-06
Updates for fixing console errors

trancjessica:
### Description
This change fixes some of the console errors that we have. Specifically ariaLabels are swapped to aria-label, and some of the attributes are set to lowercase instead of camel case (so that they can be picked up in the DOM)

### Related ticket(s)

MDCT-3036

---
### How to test

1. ./dev local
2. downloaded a screen reader
3. tabbed through the homepage and some of the work plan to validate that the screen reader could read the things on the page

https://github.com/Enterprise-CMCS/macpro-mdct-mfp/assets/9246062/3ba213c2-dedb-4e4f-ad0e-203a102060c3

### Important updates

---
### Author checklist

---


services/ui-src/src/verbiage/pages/accordion.ts
```diff
@@ -85,7 +85,7 @@ export default {
           props: {
             href: "https://www.ecfr.gov/current/title-42/chapter-IV/subchapter-C/part-438/subpart-A/section-438.8#p-438.8(k)",
             target: "_blank",
-            ariaLabel: "Link opens in new tab",
+            "aria-label": "Link opens in new tab",
```
gmrabian, line 88: I _think_ for these labels you want to include the visible text as well. 
`"aria-label": "Help File (link opens in new tab)"`
etc. 
When I tried this out with VoiceOver it only read the aria label (Link opens in new tab) and not the text (Help File), so I think the label needs both pieces of info. Maybe a different screen reader had a different result though?

services/ui-src/src/components/cards/TemplateCard.tsx
```diff
@@ -78,7 +78,7 @@ export const TemplateCard = ({
 };
 
 interface Props {
-  templateName: string;
+  templatename: string;
```
gmrabian, line 81: I'm not sure I understand this change 🤔 
what does having `templatename` in the DOM do? Also I compared main and this branch and it's all lowercase in both DOMs

services/ui-src/src/components/reports/ModalOverlayReportPage.tsx
```diff
@@ -199,7 +199,7 @@ export const ModalOverlayReportPage = ({
         <Box sx={sx.content}>
           <ReportPageIntro
             text={verbiage.intro}
-            reportType={report?.reportType}
+            reporttype={report?.reportType}
```
gmrabian, line 202: same question here about advantage of having this all lowercase in DOM

services/ui-src/src/components/cards/TemplateCard.tsx
trancjessica: I can provide some more context! :) 

This change was specifically to fix this error: 

```
Warning: React does not recognize the `templatename` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `templatename` instead. If you accidentally passed it from a parent component, remove it from the DOM element.

```
> Also I compared main and this branch and it's all lowercase in both DOMs

Oh hmmm, that is curious. Did it just get added anyways?

services/ui-src/src/verbiage/pages/accordion.ts
trancjessica: ohh, nice catch! 🌿

I tested this on Google Chrome using [this Screen Reader](https://chrome.google.com/webstore/detail/screen-reader/kgejglhpjiefppelpmljglcjbhoiplfn/related) and it does indeed just say "Link opens in new tab" when clicking "User Guide" or "Help File". I will add another update that includes the visible text!

services/ui-src/src/components/cards/TemplateCard.tsx
gmrabian: Aha so I just tested in the deployed dev instance (https://mdctmfpdev.cms.gov) and compared to local instance on main and the error only shows up when running locally 🤔 I did not see it in the deployed environment.

gmrabian: So `templateName` is passed in from HomePage but it's never actually used in TemplateCard. Could we remove that prop and then we won't have this error or the extra code? I think right now the attribute is getting spread into the Flex via ...props, but I don't think we need it in the DOM anyway. It seems like verbiage handles all of the differences between each template card type

services/ui-src/src/components/reports/ModalOverlayReportPage.tsx
gmrabian: It looks like the same thing is happening here where `reportType/reporttype` is never used as a property in ReportPageIntro so I would just remove it here

trancjessica: Yess! I can remove `reportType` 🙂

services/ui-src/src/utils/testing/generalInfoMockForm.tsx
```diff
@@ -20,7 +20,7 @@ export const mockGeneralInfoReportPageVerbiage = {
       props: {
         href: "https://www.govinfo.gov/content/pkg/PLAW-109publ171/pdf/PLAW-109publ171.pdf",
         target: "_blank",
-        ariaLabel: "Link opens in new tab",
+        "aria-label": "Link opens in new tab",
```
trancjessica, line 23: oops, looks like i missed this one. will add another commit

trancjessica: updating this file for now, in case we are using it in the future

services/app-api/forms/wp.json
```diff
@@ -27,7 +27,7 @@
               "props": {
                 "href": "https://www.govinfo.gov/content/pkg/PLAW-109publ171/pdf/PLAW-109publ171.pdf",
                 "target": "_blank",
-                "ariaLabel": "Link opens in new tab"
+                "aria-label": "User Guide and Help File (Link opens in new tab)"
```
gmrabian, line 30: This doesn't line up with the visual content 🤔 
should probably be something like
`6071(a)(1) of the Deficit Reduction Act (DRA) (Link opens in new tab)`

services/app-api/forms/wp.json
```diff
@@ -888,12 +888,12 @@
                   "props": {
                     "href": "https://www.medicaid.gov/sites/default/files/2023-01/MFP-Technical-Assistance-Brief.pdf",
                     "target": "_blank",
-                    "ariaLabel": "Link opens in new tab"
+                    "aria-label": "Link opens in new tab"
```
gmrabian, line 891: Need content in here
`“Using Data to Improve Money Follows the Person Program Performance.“ (Link opens in new tab)`

services/ui-src/src/utils/testing/generalInfoMockForm.tsx
```diff
@@ -20,7 +20,7 @@ export const mockGeneralInfoReportPageVerbiage = {
       props: {
         href: "https://www.govinfo.gov/content/pkg/PLAW-109publ171/pdf/PLAW-109publ171.pdf",
         target: "_blank",
-        ariaLabel: "Link opens in new tab",
+        "aria-label": "User Guide and Help File (Link opens in new tab)",
```
gmrabian, line 23: doesn't match content

gmrabian (changes requested):
Your changes didn't cause this but I noticed this issue/typo on the SAR dashboard where User Guide and Help File are newlined and smooshed together. Maybe this is being fixed elsewhere, just thought I'd call it out.

Screenshot 2023-10-10 at 2 26 55 PM

services/app-api/forms/wp.json
trancjessica: oh whoops! I don't know how I mixed that up, to be honest.... 😅

trancjessica:
> Good changes so far! Thanks for responding to my feedback. 😄
> 
> Just a few more below:
> 
> Your changes didn't cause this but I noticed this issue/typo on the SAR dashboard where User Guide and Help File are newlined and smooshed together. Maybe this is being fixed elsewhere, just thought I'd call it out.
> 
> Screenshot 2023-10-10 at 2 26 55 PM
> 
> from the `sar-dashboard` verbiage file

Ah. Nice catch. I can fix this!

gmrabian (approved):
lgtm!

services/app-api/forms/wp.json
```diff
@@ -891,55 +891,21 @@
                   ],
                   "text": ""
                 },
-                {
```
trancjessica: hmmm, I don't think this merge resolution did what I wanted it to do... i will be making another update

gmrabian (approved):

gmrabian (approved):

PR #146, opened 2023-10-10
Code Refactor - Repeatable ChoiceList Options - 2984 

keeysnc:
### Description
[ This is a code refactor of the 2984 ticket ] 
Goal: New "Other" Target Populations should be displayed on the 'Define Initiative'  but the field data shouldn't be mutated. 

The 'Define Initiative' page now reflects “other target populations” that are entered in the 'Benchmark Projections' section. There's only target population logic ( repeatableTargetPopulationChoices function in form.ts) to fufill the ticket but the *getRepeatableChoiceLists* function can be used as a utility to create other repeat other choice lists.

Screenshot 2023-10-04 at 2 07 47 PM

Screenshot 2023-10-04 at 2 08 08 PM

### Related ticket(s)
MDCT-[https://qmacbis.atlassian.net/browse/MDCT-2813]

https://github.com/Enterprise-CMCS/macpro-mdct-mfp/assets/14514294/5f6c9030-45f7-403b-a802-f4a2f1654fbe

---
### Author checklist

---


britt-mo (approved):
Looked good in the browser

services/ui-src/src/components/forms/Form.tsx
```diff
@@ -48,7 +48,35 @@ export const Form = ({
...
+  const updateRenderFields = (fields: (FormField | FormLayoutElement)[]) => {
+    const updatedTargetPopulationChoices = report?.fieldData?.targetPopulations;
+
+    const formatChoiceList = updatedTargetPopulationChoices?.map(
+      (field: EntityShape, key: string) => {
+        return {
+          checked: false,
+          id: field.id,
+          label:
+            parseInt(key) < 4
```
cassandradanger, line 61: what does the number 4 represent here? Might be beneficial to assign it to a variable to name it, as opposed to introducing a magic number here.

cassandradanger: if it's to check if it's one of those 4 default options...i wonder if we can check to see if it's actually one of the 4 default target pops, as opposed to one of the first 4 items in the array.  Maybe create (or utilize, if it already exists) an array with those 4 options and check if the string exists in that array

keeysnc: Good point. It's for the 4 default target populations that are already generated. I can make that a variable and use the initial array.

services/ui-src/src/components/forms/Form.tsx
```diff
@@ -51,14 +51,15 @@ export const Form = ({
 
   const updateRenderFields = (fields: (FormField | FormLayoutElement)[]) => {
     const updatedTargetPopulationChoices = report?.fieldData?.targetPopulations;
+    const defaultTargetPopulationCount = 4;
```
cassandradanger, line 54: my concern is, what happens if that list is ever altered? How would we know to update this number

ntsummers1: Yeah.. I feel you on this. Maybe we can get around this when we are attempting to add the other target population? Like the only reason we need the 4 is to mark whether its going to be one of the original ones or not. So if we add it during entity creation then we can remove this hard coded 4

keeysnc: @ntsummers1 The default target populations have a require true property on them. I used that in the current push, the "Other" Target populations dont generate the property so that might work to distinguish them.

ntsummers1: Yeah, I'm cool with that!

ntsummers1 (approved):
LGTM!

britt-mo (approved):

PR #150, opened 2023-10-11
make choice list fields correctly hydrate when re-opening a modal or drawer

britt-mo:
### Description
The nested fields in our choice lists weren't correctly hydrating when a user selected a different radio option, exited the modal or drawer, and then re-entered it. This code fixes that issue. It's not pretty but the plan is to let Zustand take care of state management issues like these so this code won't be around long-term.

### Related ticket(s)

[CMDCT-2](https://jiraent.cms.gov/browse/CMDCT-2)

---
### How to test
Go into a drawer that has radio options (like the initiatives). 
1. See that it saves correctly and hydrates correctly when you re-enter
2. Choose a different selection, go back to the original selection, see that all data has been cleared out.
Don't save, just exit. Re-enter and see that the original saved selection has correctly hydrated.

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/components/drawers/Drawer.tsx
+++ b/services/ui-src/src/components/drawers/Drawer.tsx
@@ -1,4 +1,3 @@
-import { MouseEventHandler } from "react";
 // components
 import {
   Button,
@@ -45,7 +44,7 @@ export const Drawer = ({
             sx={sx.drawerCloseButton}
             leftIcon={<CloseIcon />}
             variant="link"
-            onClick={onClose as MouseEventHandler}
+            onClick={onClose}
           >
             Close
           </Button>
@@ -64,7 +63,7 @@ interface Props {
   };
   drawerDisclosure: {
     isOpen: boolean;
-    onClose: Function;
+    onClose: any;
   };
   selectedEntity?: string;
   [key: string]: any;
--- a/services/ui-src/src/components/drawers/Drawer.tsx
+++ b/services/ui-src/src/components/drawers/Drawer.tsx
@@ -10,6 +10,7 @@ import {
   Box,
 } from "@chakra-ui/react";
 import { CloseIcon } from "@cmsgov/design-system";
+import { MouseEventHandler } from "react";
 // utils
 import { AnyObject, CustomHtmlElement } from "types";
 import { makeMediaQueryClasses, parseCustomHtml } from "utils";
@@ -44,7 +45,7 @@ export const Drawer = ({
             sx={sx.drawerCloseButton}
             leftIcon={<CloseIcon />}
             variant="link"
-            onClick={onClose}
+            onClick={onClose as MouseEventHandler}
           >
             Close
           </Button>
@@ -63,7 +64,7 @@ interface Props {
   };
   drawerDisclosure: {
     isOpen: boolean;
-    onClose: any;
+    onClose: Function;
   };
   selectedEntity?: string;
   [key: string]: any;
--- a/services/ui-src/src/utils/forms/forms.ts
+++ b/services/ui-src/src/utils/forms/forms.ts
@@ -19,6 +19,7 @@ import { DateField } from "components/fields/DateField";
 import { DropdownField } from "components/fields/DropdownField";
 import { DynamicField } from "components/fields/DynamicField";
 import { NumberField } from "components/fields/NumberField";
+import { ChoiceList } from "@cmsgov/design-system";
 
 // return created elements from provided fields
 export const formFieldFactory = (
--- a/services/ui-src/src/utils/forms/forms.ts
+++ b/services/ui-src/src/utils/forms/forms.ts
@@ -252,3 +252,28 @@ export const getRepeatableChoiceLists = (
     return fields;
   }
 };
+
+/*
+ * This function resets the 'clear' prop on each field after a ChoiceListField calls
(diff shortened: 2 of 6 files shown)


services/ui-src/src/components/drawers/Drawer.tsx
```diff
@@ -64,7 +63,7 @@ interface Props {
   };
   drawerDisclosure: {
     isOpen: boolean;
-    onClose: Function;
+    onClose: any;
```
gmrabian, line 66: was the change to `any` just to cleanup the `MouseEventHandler`?

services/ui-src/src/utils/forms/forms.ts
```diff
@@ -203,3 +204,76 @@ export const flattenFormFields = (formFields: FormField[]): FormField[] => {
...
+  );
+
+  // add more conditional logic for other choice lists
+  if (isRepeatableTargetPopulations.length > 0) {
+    repeatableTargetPopulationChoices(fields, choiceList);
+    return fields;
+  } else {
+    return fields;
+  }
+};
```
gmrabian, line 254: I don't see these two methods used in this PR

services/ui-src/src/utils/forms/forms.ts
```diff
@@ -203,3 +204,76 @@ export const flattenFormFields = (formFields: FormField[]): FormField[] => {
...
+        });
+        field.props = { ...field.props, clear: false };
+        resetClearProp(field.props?.choices);
+        break;
+      default:
+        field.props = { ...field.props, clear: false };
+        break;
+    }
+  });
+};
```
gmrabian, line 279: I wanted to learn more about this so I looked at the companion method: `ChoiceListField` -> `clearUncheckedNestedFields`

At the end of each case there I added this line and it seemed to work the same as this additional function. Is there something timing related that would make that an issue?
`child.props = { ...child.props, clear: false };`



```
        switch (child.type) {
            case "radio":
            case "checkbox":
              if (child.props?.choices) {
                child.props.choices.forEach((choice: FieldChoice) => {
                  choice.checked = false;
                });
                child.props = { ...child.props, clear: true };
                form.setValue(child.id, []);
                form.unregister(child.id);
                clearUncheckedNestedFields(child.props.choices);
+               child.props = { ...child.props, clear: false };
              }
              break;
            default:
              child.props = { ...child.props, clear: true };
              form.setValue(child.id, "");
              form.unregister(child.id);
+             child.props = { ...child.props, clear: false };
              break;
          }
```

services/ui-src/src/utils/forms/forms.ts
britt-mo: this is from a merge conflict I had

services/ui-src/src/utils/forms/forms.ts
```diff
@@ -203,3 +204,76 @@ export const flattenFormFields = (formFields: FormField[]): FormField[] => {
...
   return flattenedFields;
 };
+
+// repeatable target population choice list
+const repeatableTargetPopulationChoices = (
+  fields: (FormField | FormLayoutElement)[],
+  choiceList: any
+) => {
+  const createLabel = (field: any) => {
+    if (choiceList.indexOf(field) >= 4) {
```
trancjessica, line 214: [nit, non-blocking]: would you be open to assigning 4 to a variable so we don't have magic strings and to make it more clear that 4 is the number of "default" target populations?

services/ui-src/src/utils/forms/forms.ts
britt-mo: Did this also clear out the nested user input when going from `yes` to `no` and back to `yes`?

services/ui-src/src/utils/forms/forms.ts
britt-mo: yeah, this code is actually from Nate's work, just also here because of a merge conflict. I should adjust this PR

services/ui-src/src/components/drawers/Drawer.tsx
britt-mo: yeah we had been modifying the function call there, just reverted it though

services/ui-src/src/utils/forms/forms.ts
ntsummers1: Holy.. If that works because the `form.setValue()` call is causing a re-render of the fields and thus is letting you call child.props = {...child.props, clear: false }; just after... I'd be so happy

ntsummers1: Think it might not cover the case though that Britt just mentioned

gmrabian: @britt-mo if you have not saved anything yet, and you select `yes`, fill all the fields out, select `no`, then select `yes` again, it does clear all the fields.
However if you select `yes`, fill out the fields, save it, re-enter, then `no` then `yes` it does _not_ clear them. So I think my suggestion doesn't cover the case you specified.
That said, why would `clearUncheckedNestedFields` not handle that 🤔 
I'm ok with approving now that I see the benefit this provides. Just poking at it to see where the real breakdown is.

britt-mo: @gmrabian I'm sure there's a cleaner way to get this working lol but at the same time I know all this code will get removed once we're using Zustand so I'm inclined to move forward with this bandaid solution.

britt-mo:
Closing this PR for a cleaned up version

PR #156, opened 2023-10-13
fix bug involving nested fields correctly clearing out and also repopulating with data

britt-mo:
### Description
The nested fields in our choice lists weren't correctly hydrating when a user selected a different radio option, exited the modal or drawer, and then re-entered it. This code fixes that issue. It's not pretty but the plan is to let Zustand take care of state management issues like these so this code won't be around long-term.

This is a cleaned up version of [this closed PR ](https://github.com/Enterprise-CMCS/macpro-mdct-mfp/pull/150) that had some extra code from a merge conflict.

### Related ticket(s)

https://jiraent.cms.gov/browse/CMDCT-2

---
### How to test
Go into a drawer that has radio options (like the initiatives).

1. See that it saves correctly and hydrates correctly when you re-enter
2. Choose a different selection, go back to the original selection, see that all data has been cleared out.
    Don't save, just exit. Re-enter and see that the original saved selection has correctly hydrated.

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/components/reports/ModalDrawerReportPage.tsx
+++ b/services/ui-src/src/components/reports/ModalDrawerReportPage.tsx
@@ -39,6 +39,7 @@ import {
   getEntriesToClear,
   setClearedEntriesToDefaultValue,
   entityWasUpdated,
+  resetClearProp,
 } from "utils";
 
 export const ModalDrawerReportPage = ({ route, validateOnRender }: Props) => {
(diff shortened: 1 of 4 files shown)


gmrabian (approved):
Verified locally

trancjessica (approved):
tested locally and verified that the fields are repopulating correctly when data is changed

PR #167, opened 2023-10-16
set up route

cassandradanger:
### Description

PR to set up the /export routes for WP and SAR PDFs

this does not cover any PDF creation or downloading, just the route

### Related ticket(s)

CMDCT-207

---
### How to test

- log in as user or admin and skip to the Review and Submit page of the SAR or the WP
- click Review PDF
- a new tab should open for the /export route and the component should say what report the PDF page is for

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/components/app/AppRoutes.tsx
+++ b/services/ui-src/src/components/app/AppRoutes.tsx
@@ -11,6 +11,7 @@ import {
   ProfilePage,
   ReportPageWrapper,
   ReviewSubmitPage,
+  ExportedReportPage,
 } from "components";
 // utils
 import { ScrollToTopComponent, useStore } from "utils";
@@ -41,16 +42,24 @@ export const AppRoutes = () => {
           <Route path="*" element={<NotFoundPage />} />
           {/* MFP Report Routes */}
           {wpReport && (
-            <Route
-              path="/wp"
-              element={<DashboardPage reportType={ReportType.WP} />}
-            />
+            <>
+              <Route
+                path="/wp"
+                element={<DashboardPage reportType={ReportType.WP} />}
+              />
+              {/* TODO check for report access w userReportAccess */}
+              <Route path="/wp/export" element={<ExportedReportPage />} />
+            </>
           )}
           {sarReport && (
-            <Route
-              path="/sar"
-              element={<DashboardPage reportType={ReportType.SAR} />}
-            />
+            <>
+              <Route
+                path="/sar"
+                element={<DashboardPage reportType={ReportType.SAR} />}
+              />
+              {/* TODO check for report access w userReportAccess */}
+              <Route path="/sar/export" element={<ExportedReportPage />} />
+            </>
           )}
           {/* General Report Routes */}
           {report && (
--- a/services/ui-src/src/components/app/AppRoutes.tsx
+++ b/services/ui-src/src/components/app/AppRoutes.tsx
@@ -19,8 +19,12 @@ import { ScrollToTopComponent, useStore } from "utils";
 import { ReportRoute, ReportType } from "types";
 
 export const AppRoutes = () => {
-  const { userIsAdmin } = useStore().user ?? {};
+  const { userIsAdmin, userReports } = useStore().user ?? {};
   const { report } = useStore();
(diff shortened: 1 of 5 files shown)


services/ui-src/src/types/users.ts
```diff
@@ -15,6 +15,7 @@ export interface MFPUser {
   full_name: string;
   state?: string;
   userRole?: string;
+  userReports?: string[] | undefined;
```
gmrabian, line 18: will this ever be undefined if we are setting the [WP, SAR] array when reports is undefined?

services/ui-src/src/components/app/AppRoutes.tsx
```diff
@@ -11,15 +11,20 @@ import {
...
+  ExportedReportPage,
 } from "components";
 // utils
 import { ScrollToTopComponent, useStore } from "utils";
 // types
 import { ReportRoute, ReportType } from "types";
 
 export const AppRoutes = () => {
-  const { userIsAdmin } = useStore().user ?? {};
+  const { userIsAdmin, userReports } = useStore().user ?? {};
```
gmrabian, line 22: I don't understand the purpose of `userReports` 🤔

services/ui-src/src/components/app/AppRoutes.tsx
```diff
@@ -41,16 +46,40 @@ export const AppRoutes = () => {
...
+                path="/wp"
+                element={<DashboardPage reportType={ReportType.WP} />}
+              />
+              <Route
+                path="/wp/export"
+                element={
+                  userReportAccess["WP"] ? (
+                    <ExportedReportPage />
+                  ) : (
+                    <Navigate to="/" />
```
gmrabian, line 60: I tried navigating straight to `/wp/export` as a state user and as an admin user, both within a report and outside. in all cases I was able to get there

gmrabian (commented):
curious about some details here. PDF review page shows up as expected when clicking that link in the reports.

services/ui-src/src/components/app/AppRoutes.tsx
braxex: Pretty sure we don't need `userReports` because all users have access to both reports in MFP (as opposed to MCR where in the past we've had only some users have access to certain reports).

services/ui-src/src/utils/auth/UserProvider.tsx
```diff
@@ -74,6 +74,8 @@ export const UserProvider = ({ children }: Props) => {
       const userRole = cms_role.split(",").find((r) => r.includes("mdctmfp"));
       const state = payload["custom:cms_state"] as string | undefined;
       const full_name = [given_name, " ", family_name].join("");
+      const reports = payload["custom:reports"] as string | undefined;
+      const userReports = !reports ? ["WP", "SAR"] : reports.split(",");
```
braxex, line 78: Note: Can delete this stuff because we're not restricting any users from SAR or WP.

services/ui-src/src/utils/auth/UserProvider.tsx
```diff
@@ -88,6 +90,7 @@ export const UserProvider = ({ children }: Props) => {
         full_name,
         userRole,
         state,
+        userReports,
```
braxex, line 93: Same note as above.

services/ui-src/src/types/users.ts
braxex: Don't need to add this because we're not restricting user report access for MFP.

braxex (commented):
Just a couple notes -- all related to us not needing to restrict user access to reports in MFP, so we don't need `userReports`.

services/ui-src/src/components/app/AppRoutes.tsx
braxex: yeah that’s prob something we should patch up in both — if there’s not a currently selected report, redirect user back to dashboard. i’d say let’s make a new ticket for it — one for MFP, one for MCR.

gmrabian (approved):

ajaitasaini (approved):
works for both WP and SAR!

PR #168, opened 2023-10-17
[CMDCT-159]: SAR Form Creation Cleanup Part 1

ntsummers1:
### Description

- Abstracts out code during the creation process in the api handler
- Adds a TON of comments to help the reader follow along
- Updates the value sent for the finalSar question to be in the proper Choice type format

To best read this PR, I recommend you pull it down and read the create.ts file from top to bottom. That'll be much easier to understand and review than if you try and use the Files Changed mode. I had to move some code around to make it read better so you might lose track of things or wonder why I'm deleting them when really its added back later on or further up. You can also try to read it by clicking through each commit. I tried to package everything so that the commit message makes sense for what I changed. I recommend doing that after reading create.ts separately.

### Related ticket(s)

CMDCT-159

---
### How to test

- Create a work plan
- Create a SAR
- See that you are unable to create an additional SAR (Assuming you only created the one WP)
- Check the network tab and refresh the page on the dashboard page for SAR and Workplan
- Note that there should be a respective field called sar or workplan in the response with the id of the SAR or WP its connected to
- Create 2 work plans
- Edit a field in both work plans but make those values different (So you can tell the work plans apart)
- Create a SAR and see that it built it of the third work plan you just created, not the second or first one.
- Create a SAR and see that it was built off the second work plan you created.
- Try editing a SAR, note that your answers should all be correct
- See that you can't edit a WP too

---
### Author checklist


services/ui-src/src/components/reports/ReportProvider.tsx
```diff
@@ -121,33 +122,32 @@ export const ReportProvider = ({ children }: Props) => {
         selectedState
       );
 
-      if (workPlanSubmissions?.length !== sarSubmissions?.length) {
-        let lastFoundSubmission: ReportMetadataShape | undefined = undefined;
-        workPlanSubmissions.forEach((submission: ReportMetadataShape) => {
+      let lastFoundSubmission: ReportMetadataShape | undefined = undefined;
```
ntsummers1, line 125: This will be abstracted in SAR Form Creation Cleanup Part 2, so please don't mind that its without comments and living large for now :D

services/app-api/handlers/reports/create.ts
```diff
@@ -101,91 +109,78 @@ export const createReport = handler(
...
+
+    /*
+     * Begin Section - If creating a SAR Submission, find the last Work Plan created that hasn't been used
+     * to create a different SAR and attach all of its fieldData to the SAR Submissions FieldData
+     */
+    const {
+      workPlanMetadata,
+      workPlanFieldData,
+    }:
+      | {
```
trancjessica, line 143: is it possible to simplify this to: 

```
const {
      workPlanMetadata,
      workPlanFieldData,
    }:
      {
          workPlanMetadata?: ReportMetadataShape;
          workPlanFieldData?: AnyObject;
        } =
      reportType === ReportType.SAR
        ? await getLastCreatedWorkPlan(event, _context, state)
        : { workPlanMetadata: undefined, workPlanFieldData: undefined };
```

or does it get wonky and think `?` means null instead of optionally undefined?

services/ui-src/src/types/reportContext.ts
```diff
@@ -9,21 +9,29 @@ export interface ReportKeys {
...
   dueDate: number;
   reportPeriod: number;
-  finalSar?: boolean;
+  reportYear: number;
+  // Admin Use Cases
+  archived?: boolean;
+  locked?: boolean;
+  submissionCount?: number;
+  // Connection to let Sar forms know what Work Plan they use to copy data from and vice versa
+  sar?: string;
```
trancjessica, line 31: [very small nit, non-blocking]: i feel like `sar` and `workPlan` might get misinterpreted out of context.  I wonder if it would be more clear if we called these something like `associatedSar` or `associatedWorkPlan`?

trancjessica (dismissed):
tested on my machine. sar creation functionality still works! just a few nits, but non-blocking :D ship it 🚀

services/app-api/handlers/reports/create.ts
ntsummers1: Your so right, thank you. Fixed!

services/app-api/handlers/reports/create.ts
```diff
@@ -92,6 +94,12 @@ export const createReport = handler(
       logger.error(err, "Error getting or creating template");
       throw err;
     }
+    // End Section - Getting/Creating newest Form Template based on reportType
```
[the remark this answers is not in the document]
ntsummers1: 😄

trancjessica (approved):
lgtm! 🙌

services/app-api/handlers/reports/create.ts
```diff
@@ -17,27 +17,29 @@ import {
...
   UserRoles,
 } from "../../utils/types";
 import { getOrCreateFormTemplate } from "../../utils/formTemplates/formTemplates";
 import { logger } from "../../utils/logging";
-import { fetchReport, fetchReportsByState } from "./fetch";
 import { APIGatewayProxyEvent } from "aws-lambda";
 
 export const createReport = handler(
   async (event: APIGatewayProxyEvent, _context) => {
+    // If user who is calling for report creation isn't a state user or state rep, return Unautorized
```
britt-mo, line 42: if you're gonna make more changes anyway, nit: 'unauthorized'

services/app-api/handlers/reports/create.ts
```diff
@@ -59,30 +61,30 @@ export const createReport = handler(
     }
 
     const { state, reportType } = event.pathParameters;
+
+    // If that string they passed in the state paramater isn't a real US State, return No_Key
```
britt-mo, line 65: Could also remove this comment, I think that the code is easily understandable here

services/app-api/handlers/reports/create.ts
```diff
@@ -214,14 +206,14 @@ export const createReport = handler(
         body: error.INVALID_DATA,
       };
     }
-    const currentDate = Date.now();
 
+    // Begin Section - Create DyanmoDB record.
```
britt-mo, line 210: nit: 'dynamo'

britt-mo (approved):
just some nits, looks good.

services/app-api/handlers/reports/create.ts
ntsummers1: Will do!

services/app-api/handlers/reports/create.ts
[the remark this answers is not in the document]
ntsummers1: 🙌

services/app-api/handlers/reports/create.ts
ntsummers1: Will do!

trancjessica (approved):
tested again on my machine. sar creation looks good!

trancjessica (approved):

services/app-api/utils/other/other.ts
```diff
@@ -18,3 +25,80 @@ export const createReportName = (
...
+       * ...then do one of two things: if there are multiple work plans that meet this criteria,
+       * grab the one that was created most recently and return that as our work plan to
+       * copy from...
+       */
+      if (
+        lastCreatedWorkPlan &&
+        workPlan.createdAt > lastCreatedWorkPlan?.createdAt
+      ) {
+        lastCreatedWorkPlan = workPlan;
+      } else if (!lastCreatedWorkPlan) {
```
gmrabian, line 55: suggestion: could this be simplified to 

`if (
        !lastCreatedWorkPlan ||
        workPlan.createdAt > lastCreatedWorkPlan?.createdAt
      )`

gmrabian (commented):
Create WP was disabled after creating only 1 actually, regardless of SARs

PR #176, opened 2023-10-18
CMDCT-159 Update due date calculation

trancjessica:
### Description

This change 
- adds a method to the create API to calculate the due date for a given **Work Plan** or **SAR** report
- removes the **calculateDueDate** method from the `ui-src` `time` file
- adds tests to verify if a given year is a **leap year** 
- adds tests to verify the due date is calculate correctly for a **Work Plan** or a **SAR** report created in **Period 1** or **Period 2**

### Related ticket(s)

CMDCT-159

Due Date Logic: 
 * WP due date is May 1 for Period 1, and Nov 1 for Period 2
 * SAR due date for Period 1{*}: 60 days from June 30, due date is Aug 29.
 * SAR due date for Period 2{*}: 60 days from December 31(for a non leap year Mar 1, for leap years it’s Feb 29).
 
---
### How to test

1. login as state user
2. create work plan
3. check that the due date is correct

https://github.com/Enterprise-CMCS/macpro-mdct-mfp/assets/9246062/023e2859-5865-43c7-b28c-2c55f494a8f0

Note: I am not sure why my SAR report looks a bit jank and doesn't have a due date on the frontend. Perhaps it is because the WP is not actually completed? Happy to put up more commits if I need to change something to fix this

UPDATED VIDEO!! 😄

https://github.com/Enterprise-CMCS/macpro-mdct-mfp/assets/9246062/874e012c-c7ac-472c-9230-be3805906a25

### Important updates

---
### Author checklist

- [X] I have added [thorough](https://shorturl.at/aejkF) tests, if necessary
- [X] I have updated relevant documentation, if necessary
---


services/app-api/handlers/reports/create.ts
```diff
@@ -242,7 +248,12 @@ export const createReport = handler(
           workPlanMetadata
         ),
         reportYear,
-        reportPeriod: calculatePeriod(currentDate, workPlanMetadata),
+        reportPeriod: reportPeriod,
```
gmrabian, line 251: tip: since the key and value here have the same name you can just write `reportPeriod,` like how the line above is just `reportYear,`!

services/app-api/utils/time/time.ts
```diff
@@ -21,18 +21,80 @@ export const convertDateUtcToEt = (date: number): string => {
...
+  reportType: ReportType
+) => {
+  let date = new Date();
+  if (reportType == ReportType.WP) {
+    reportPeriod == "1"
+      ? (date = new Date(currentYear, 4, 1))
+      : (date = new Date(currentYear, 10, 1));
+  }
+  if (reportType == ReportType.SAR) {
+    if (reportPeriod == "2") {
```
gmrabian, line 87: is the loose equals (double instead of triple =) intentional here? like for reportPeriod being an integer vs. a string?
[docs on loose and strict equals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness)

services/app-api/utils/time/time.ts
```diff
@@ -21,18 +21,80 @@ export const convertDateUtcToEt = (date: number): string => {
...
+  var year = date.getFullYear();
+
+  var month = (1 + date.getMonth()).toString();
+  month = month.length > 1 ? month : "0" + month;
+
+  var day = date.getDate().toString();
+  day = day.length > 1 ? day : "0" + day;
+
+  return month + "/" + day + "/" + year;
+};
```
gmrabian, line 63: Some thoughts:
1. Is `var` making a comeback? 😵 
2. Suggestion: string interpolation (combining with backticks) instead of concatenation (with pluses)
  \`${month}/${day}/${year}\` etc.

services/app-api/utils/time/time.ts
```diff
@@ -21,18 +21,80 @@ export const convertDateUtcToEt = (date: number): string => {
...
 /*
- * Calculates the period given a due date.
+ * Calculates the period given the current date.
  * The periods are defined as follows:
  *     Period 1 is from 01/01 to 06/30.
  *     Period 2 is from 07/01 to 12/31.
  */
-export const calculatePeriod = (dueDate: number, workPlan?: ReportMetadata) => {
+export const calculatePeriod = (
+  currentDate: number,
```
gmrabian, line 30: currentDate seems like a funny thing to pass in when you could just use Date.now() 🤔 
I know we get Date.now() in create.ts so we can pass it in easily, just thinking out loud. Does it hurt to calculate it here as well?

gmrabian (commented):
Lovely tests ✨ 😁 

The rest is my brain sometimes side-tracking, sometimes exploding. Feel free to enjoy or ignore at your leisure 🙃

services/app-api/utils/time/time.ts
gmrabian: There also seems to be some funny logic in general around dates and years sometimes be calculated and sometimes coming from metadata. Perhaps a larger thing to think about (or totally ignore 👍)

services/app-api/utils/time/time.ts
trancjessica: 1. Is `var` making a comeback? Not intentionally 😅 I can use `let` instead 😄
2. good suggestion!! i will revise this with the backticks

services/app-api/utils/time/time.ts
trancjessica: > There also seems to be some funny logic in general around dates and years sometimes be calculated and sometimes coming from metadata. Perhaps a larger thing to think about (or totally ignore 👍)

ohhh hmmm very good point.... I will have to think about this one some more

> currentDate seems like a funny thing to pass in when you could just use Date.now() 🤔
I know we get Date.now() in create.ts so we can pass it in easily, just thinking out loud. Does it hurt to calculate it here as well?

true. we could just calculate `Date.now()` here. also it looks like we only need the month here? so maybe we could pass in the month from create.ts? lots of paths here 🤔

services/app-api/handlers/reports/create.ts
trancjessica: ohhh nice! that's good to know!

services/app-api/utils/time/time.ts
trancjessica: it is not intentional 😅 reading the doc 😄

trancjessica: period is meant to be a number :D so I will update this to be a strict equals and update the `"1"` and `"2"` to be numbers :)

services/app-api/utils/time/time.ts
```diff
@@ -21,18 +21,80 @@ export const convertDateUtcToEt = (date: number): string => {
...
   return period.toString();
 };
 
+/**
+ * Calculates if the given year is a leap year
+ * @param year The year.
+ * @returns if the given year is a leap year.
+ */
+export const isLeapYear = (year: number) => {
+  return (year % 4 == 0 && year % 100 != 0) || year % 400 == 0;
```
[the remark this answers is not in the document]
trancjessica: I think strict equals is probably better here? since the parameter `year` is a `number`? in any case, it doesn't hurt to have the strict comparison here :D

services/app-api/utils/time/time.ts
trancjessica: On second thought, I like having the date as a parameter! it makes testing easier 😄 we can pass in a variety of dates and verify that the period is what we expect it to be. That becomes a bit trickier to do if `calculatePeriod` computes `Date.now()` itself

britt-mo:
Do you have the latest from main in this PR?

trancjessica:
> Do you have the latest from main in this PR?

ah, i dont think i do! i can pull the latest changes from main

services/app-api/utils/time/time.ts
gmrabian: true enough!

services/app-api/utils/time/time.ts
```diff
@@ -21,19 +21,78 @@ export const convertDateUtcToEt = (date: number): string => {
...
+  reportPeriod: number,
+  reportType: ReportType
+) => {
+  let date = new Date();
+  if (reportType == ReportType.WP) {
+    reportPeriod === 1
+      ? (date = new Date(currentYear, 4, 1))
+      : (date = new Date(currentYear, 10, 1));
+  }
+  if (reportType == ReportType.SAR) {
```
gmrabian, line 86: suggestion: I'd recommend a strict equals on reportType as well. Doesn't hurt at least!

services/app-api/handlers/reports/create.ts
```diff
@@ -232,8 +239,7 @@ export const createReport = handler(
         ),
         targetPopulations: "Older adults, PD, MH/SUD",
         reportYear,
-        dueDate: currentDate,
-        reportPeriod: calculatePeriod(currentDate, workPlanMetadata),
+        dueDate: calculateDueDate(reportYear, reportPeriod, reportType),
         associatedWorkPlan: workPlanMetadata?.id,
       },
```
gmrabian, line 244: did you mean to remove `reportPeriod`?

trancjessica: whoops! I did not. Good catch! I'll add it back in

ntsummers1 (dismissed):
Looks great!

cassandradanger (approved):
looks good!!

ntsummers1 (approved):

PR #179, opened 2023-10-19
Cmdct 3019 - add period info for R, E, T pages except for page nine

britt-mo:
### Description
Each page in the `Recruitment, Enrollment, and Transitions` section needs the reporting period displayed. The exception is page 9, so I've excluded it and that work is its own card. I put this section into its own component and then brought it into the `ReportPageIntro` component. 

Also, the design has been updated, here is a screenshot of the spec:
Screenshot 2023-10-19 at 2 38 00 PM

### Related ticket(s)

CMDCT-3019

---
### How to test
See that the reporting period text only appears in the Recruitment, Enrollment, and Transitions section, and page 9 differs. Also see that this text doesn't appear on any other pages (this component is shared across the WP and SAR.

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/components/reports/ReportPeriod.tsx
+++ b/services/ui-src/src/components/reports/ReportPeriod.tsx
@@ -1,0 +1,39 @@
+// components
+import { Box, Heading } from "@chakra-ui/react";
+// utils
+import { displayLongformPeriod } from "utils";
+import { AnyObject } from "types";
+
+export const ReportPeriod = ({ text, reportPeriod }: Props) => {
+  const { section, subsection } = text;
+  const retSection = section === "Recruitment, Enrollment, and Transitions";
+  const pageNine = subsection.includes("HCBS");
+  const currentPeriod = displayLongformPeriod(reportPeriod);
+  return (
+    <Box>
+      {retSection &&
+        (!pageNine ? (
+          <Heading as="h2" sx={sx.periodText}>
+            {currentPeriod}
+          </Heading>
+        ) : (
+          <Heading as="h2" sx={sx.periodText}>
+            This page needs a different calculation
+          </Heading>
+        ))}
+    </Box>
+  );
+};
+
+interface Props {
+  text: AnyObject;
+  [key: string]: any;
+  reportPeriod?: number;
+}
+
+const sx = {
+  periodText: {
+    marginTop: "1.5rem",
+    fontSize: "2xl",
+  },
+};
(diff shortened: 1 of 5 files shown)


trancjessica (commented):
tested locally! lgtm! one thing I would ask product about is if we want the `{ }` showing up? I worked on a ticket where the `{ }` was in the mocks, but it turns out they didn't want them on the page. They just used it to signify the values could change

britt-mo:
@trancjessica I did talk to design about that, the `{}` are supposed to display

trancjessica (dismissed):

services/ui-src/src/components/reports/ReportPeriod.tsx
```diff
@@ -0,0 +1,39 @@
...
+  return (
+    <Box>
+      {retSection &&
+        (!pageNine ? (
+          <Heading as="h2" sx={sx.periodText}>
+            {currentPeriod}
+          </Heading>
+        ) : (
+          <Heading as="h2" sx={sx.periodText}>
+            This page needs a different calculation
```
karla-vm, line 21: Is this a placeholder for now?

karla-vm (dismissed):
Nice! This component is going to be super helpful for the other RE&T pages too, awesome work.

ajaitasaini (dismissed):
seeing the changes correctly!

services/ui-src/src/components/reports/ReportPeriod.tsx
britt-mo: yep haha cause Casie said page 9 is its own card

services/ui-src/src/utils/other/time.ts
```diff
@@ -163,3 +163,12 @@ export const calculateRemainingSeconds = (expiresAt?: any) => {
   if (!expiresAt) return 0;
   return moment(expiresAt).diff(moment()) / 1000;
 };
+
+export const displayLongformPeriod = (period: number | undefined) => {
+  const currentYear = new Date().getFullYear();
```
ntsummers1, line 168: We can grab the reportYear from the metadata here as well. No need to calculate the current year!

ntsummers1: Might be important because if you review an old SAR the current year would be when you view it, not when the SAR Was actually created

britt-mo: good point

ntsummers1 (approved):
Thanks for the quick fix!

PR #199, opened 2023-11-02
start bringing over PDF components from MCR

cassandradanger:
### Description

The second step in creating the WP PDF - porting over PDF components from MCR. 

There is probably a component or two that I won't end up needing, but I'm bringing them over now, and will delete later if I don't need them, or can't repurpose them.

### Related ticket(s)

CMDCT-218

---
### How to test

You can create a workplan and then click Review PDF to be taken to the PDF export - there are a few things showing up, and hopefully no breaking errors.
You can fill out a WP and click Review PDF to see the export, which should have a little bit more info and components to see, and also no breaking errors.

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/components/export/ExportedReportFieldTable.tsx
+++ b/services/ui-src/src/components/export/ExportedReportFieldTable.tsx
@@ -1,0 +1,204 @@
+import { ReactElement } from "react";
+// components
+import { Table } from "components";
+// types, utils
+import { useStore } from "utils";
+import {
+  Choice,
+  EntityShape,
+  FieldChoice,
+  FormField,
+  StandardReportPageShape,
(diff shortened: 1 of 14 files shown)


services/ui-src/src/components/export/ExportedReportFieldTable.tsx
```diff
@@ -0,0 +1,204 @@
+import { ReactElement } from "react";
```
cassandradanger, line 1: I might not need this component, since it doesn't look like we have a ReportFieldTable - but I figure I'll bring it over and delete it if I can't repurpose it

gmrabian (approved):
export page has more stuff! some is just spilled out json, though I know this is WIP

PR #215, opened 2023-11-14
[MFP] SAR - R,E,&T - Quarter and Repeating Fields Logic

ajaitasaini:
### Related ticket(s)

CMDCT-3021

---
### How to test

- Login, create a work plan, and (optionally) add other target populations under transition benchmarks 
- Leave form, add a new SAR 
- Navigate to "Number of MFP transitions in the reporting period" under "Recruitment, Enrollment, and Transitions" 
- Confirm that you see the correct reporting period and corresponding quarters within that period, and all the target populations per quarter ([Figma here](https://www.figma.com/file/oOJMUZerbafZefLQQ33GwI/MFP_Playground?type=design&node-id=246-53705&mode=design&t=acyOhY5aHn9Qfi7R-0))
- Go to other pages under RET and confirm that they are unchanged

### Important updates

---
### Author checklist

---


services/ui-src/src/utils/reports/entities.ts
```diff
@@ -108,7 +108,7 @@ export const renderTargetPopulationFields = (
         id: field.id,
         label: field.isRequired
           ? field.transitionBenchmarks_targetPopulationName
-          : `Other: ${field.transitionBenchmarks_targetPopulationName}`,
+          : `Other: {${field.transitionBenchmarks_targetPopulationName}}`,
```
gmrabian, line 111: I'm not sure what this change does 🤔 
I tested with and without and didn't see a visual difference, though I would expect this change to add curly braces to the view

services/app-api/utils/formTemplates/formTemplates.ts
```diff
@@ -89,9 +89,116 @@ export const nextTwelveQuarters = (
...
+    workPlanMetaData
+  );
+
+  targetPopulationsByQuarter(
+    formFields,
+    fieldToRepeat,
+    2,
+    workPlanFieldData,
+    workPlanMetaData
+  );
```
gmrabian, line 188: Nit: If this method is always called twice, with quarters 1 and 2, could you just call it once and have it do both?

ajaitasaini: For now this method is only called twice, but for other pages in future implementation it'll be called a different amount of times, which is why I wanted to divide it up this way. I could put this in a loop so that there's less repeated code

gmrabian: This clears things up. I'll leave it to you, just wanted to check!

gmrabian (approved):

britt-mo (approved):
Looks good, I'd like to get this merged in before my (small) tweak to the repeating quarters fields.

PR #230, opened 2023-11-20
[PenTest] - Remove renderHTML from places it dont belong

ntsummers1:
### Description

renderHTML was being called in entityRows and allowing for user entered data to be shown. Whoops! Was an accidental copy from MCR's exportedEntityRow and has now been removed.

---
### How to test

Create a Workplan
Create a transition benchmark or initiative and pass into the textbox `<iframe src=javascript:alert(document.domain)>`
See that no alert pops up!

---
### Author checklist

---


ailZhou (approved):
looks good, thank you for fixing this!

gmrabian (approved):
⭐ good find! ⭐

PR #240, opened 2023-11-28
CMDCT-2375 [Bug] Update end date validation

trancjessica:
### Rev 2
This change fixes the End Date field to validate that the end date is after the start date. It also implements a radio button so that the user can click "No" if they don't have an end date. 

### Rev 1 [IGNORE]
Hello! This is the work I have done so far to fix the following issues: 
Define initiative, End Date field: 
1. End Date field needs to allow "N/A" as an answer. 
2. It allows an end date to be before a start date, we should try to prevent that

The changes I have committed definitely fix `#2`. However, when `N/A` is input in the field, it then disappears when the user clicks outside of the box. I would love some help if anyone has insight on why this is happening! 😄

I am including a video below to demonstrate the changes!

Side note: I do plan on adding tests once I figure out how to fully fix this bug! 

### Related ticket(s)

CMDCT-2375

---
### How to test

1. Sign in as a state user
2. Navigate to the State and Territory Specific Initiatives 
3. Create an initiative
4. Edit the initiative
5. Define the initiative
6. Enter a start date
7. Enter no end date and notice the error: `Response must be a valid date or N/A'
8. Enter an end date **before the start date** and notice the error `End date must be after start date`
9. Enter an end date **after the start date** and notice it works as expected
10. Enter an end date of `N/A` and it disappears 🤔

https://github.com/Enterprise-CMCS/macpro-mdct-mfp/assets/9246062/7186efd1-9623-4e59-bd1a-bc4368f4d302

### Important updates

---
### Author checklist

---


karla-vm (approved):
LGTM!

PR #248, opened 2023-11-30
[CMDCT-3087] Cleanup backend data of no longer applicable target populations

ntsummers1:
### Description

If a user was to speed ahead and define an initiative, then check a population as part of that initiative, and then back track and mark a population as not applicable... Well, the data would live on in the initiative. This would mean it could accidentally get passed to the SAR or a new Work Plan when its copied over. So we need to clean up after this possible scenario! And wouldn't ya know it, I've done just that in this here PR :) 

https://github.com/Enterprise-CMCS/macpro-mdct-mfp/assets/19439679/a4f66186-6e61-4f3c-8133-2e615c558bdb

### Related ticket(s)

CMDCT-3087

---
### How to test

Create a Work Plan
Create an initiative
Go to the define initiative page
Checkmark the first couple populations
Go to the Transition Benchmarks Page
Mark the Older Populations as not applicable (Or any you want, really)
See that the data you get back (By checking the preview tab and then looking at FieldData -> Initiative) and see that only the target populations that you didn't just mark as not applicable are still there

---
### Author checklist


services/app-api/utils/data/data.ts
```diff
@@ -0,0 +1,65 @@
...
+        (initiativePopulation: AnyObject) => {
+          for (const population of notApplicablePopulations) {
+            const populationName = population?.isRequired
+              ? population?.transitionBenchmarks_targetPopulationName
+              : `Other: ${population?.transitionBenchmarks_targetPopulationName}`;
+            if (populationName == initiativePopulation?.value) {
+              return false;
+            }
+          }
+          return true;
```
benmartin-coforma, line 56: THOUGHT - it may be worthwhile to map `notApplicablePopulations` into `notApplicablePopulationNames` ahead of time. That would allow you to replace this for loop with 

```ts
(initiativePopulation: Record<string, any>) =>
  notApplicablePopulationNames.includes(initiativePopulation.value)
```

services/app-api/utils/data/data.ts
benmartin-coforma, line 1: NITS - There are a few places in here with safe-access operators that aren't doing anything. Using `foo?.bar` protects you from errors when `foo` is nullish, and has no effect otherwise. So you could remove those from:
* lines 14 and 15: because `fieldData` is not null
* line 31: because `population` is not null - I assume. Unless it's possible for `targetPopulations` to include a mix of populations and null values?
* line 46: because `initiative` is not null - I assume. If it's possible for `initiatives` to include a mix of initiatives and null values, then you'll need to add more safety around the `.` on line 45.
* lines 49, 50, and 51: because `population` is not null. Even if it could have been on line 31, you've eliminated that possibility by this point.
* line 52: because `initiativePopulation` is not null - I assume. Unless it's possible for `defineInitiative_targetPopulations` to contain a mix of populations and null values.

benmartin-coforma (approved):
I have a few nits to pick, but as far as I can tell the code will work fine - and the variable naming & commenting were immensely helpful to my understanding.

I'll also note that I have not tested this.

services/app-api/utils/data/data.ts
benmartin-coforma: But even if the fieldData doesn't have targetPopulations, that's not the same as targetPopulations being an array that contains nullish entries. And that's the only thing that could make the safe-access operator necessary on `population` on line 31, for example.

Here's a branch with extra tests for more ways that data could be missing, and fewer safe-access operators, with all tests still passing. I also swapped to `.includes`, and replaced the `.map` on initiatives with a `for` (since they are being modified in-place). https://github.com/Enterprise-CMCS/macpro-mdct-mfp/compare/cmdct-3087...cmdct-3087-ben

Ultimately these are all minor differences, though. You have my approval whether you decide to make changes or not.

benmartin-coforma (dismissed):

benmartin-coforma (approved):
I heartily approve :D

PR #263, opened 2023-12-05
Disable create report button while creating is in progress

gmrabian:
### Description

When you are in the AddEditReportModal you can click the "Start new" button fast and it'll create multiple reports. Now these buttons are disabled when `submitting` is set to true, which happens at the top of the `writeReport` method the button triggers.

### Related ticket(s)

CMDCT-3101

---
### How to test

- Log in as a state user
- Go to WP dashboard and click "Start MFP Work Plan" 
- Click "Start new" as fast as you can multiple times
- Ensure only one WP gets created

To retest you can log in as an admin and archive the created WP, or restart the app.

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/components/modals/AddEditReportModal.tsx
+++ b/services/ui-src/src/components/modals/AddEditReportModal.tsx
@@ -207,7 +207,7 @@ export const AddEditReportModal = ({
         <>
           <Button
             sx={sx.copyBtn}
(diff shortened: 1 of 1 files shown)


karla-vm (approved):

britt-mo (approved):
Worked as expected in browser. Or maybe I have a very slow connection 😆

PR #269, opened 2023-12-06
Cmdct 3115 - fix ability to add and delete initiatives

britt-mo:
### Description
After submitting a workplan and having an admin unlock it, a state user couldn't delete initiatives or create or delete new nested initiatives. The issue was due to the fact that every update and delete was sending a full payload with the following 'blacklist` data, which isn't allowed after a workplan has already been submitted/is in revision:
```
    "submitterName",
    "submitterEmailAddress",
    "reportSubmissionDate",
```

### Related ticket(s)

CMDCT-3115

---
### How to test
Create a wp and submit it
login as an admin and unlock it
login as a state user and enter the unlocked workplan
verify that you can delete initiatives and create new ones at the top level and nested level (evaluation plans, funding sources)

### Important updates

---
### Author checklist

---


services/ui-src/src/components/modals/DeleteEntityModal.tsx
```diff
@@ -70,18 +70,20 @@ export const DeleteEntityModal = ({
...
     );
 
+    const tempEntities = { ...updatedEntities };
     await updateReport(reportKeys, {
       metadata: {
         lastAlteredBy: full_name,
         status: ReportStatus.IN_PROGRESS,
       },
-      fieldData: { ...updatedEntities },
+      fieldData: { [entityName]: tempEntities[entityName] },
```
ailZhou, line 86: nit: not sure if its necessary to store the updatedEntities in temp variable when its only for data retrieval and theres no modification after. you can just use it as:

`fieldData: { [entityName]: updatedEntities [entityName] },`

ailZhou (dismissed):
Tested locally and works beautifully and the logic looks sound. I don't see any issues besides one tiny nit but that's not really important. Thank you for fixing this!

cassandradanger (approved):
looks great in the browser! Thanks for removing those <br>'s in the copy as well :)

services/ui-src/src/components/modals/DeleteEntityModal.tsx
britt-mo: you were right. updated.

ailZhou (approved):

ntsummers1 (approved):

PR #284, opened 2023-12-12
Fix Define Initiative end date not populated error

gmrabian:
### Description

Report was crashing when skipping a question because it performed a check on the empty field without null handling
Video of bug in action (tab through Define Initiative and skip the radio options for end date)
https://github.com/Enterprise-CMCS/macpro-mdct-mfp/assets/57802560/fadceadb-bce1-4345-90c5-3ea540b25edf

### Related ticket(s)

CMDCT-3124

---
### How to test

Do the same steps in this branch and verify that it doesn't crash
- Create an initiative
- Go to Define Initative
- **TAB THROUGH** each option, skipping the final radio and hitting spacebar on "Save & Return"
- Verify it triggers validation on the radio instead of crashing

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/utils/forms/forms.ts
+++ b/services/ui-src/src/utils/forms/forms.ts
@@ -123,7 +123,7 @@ export const hydrateFormFields = (
      * the value of the V. Close-out initiative disabled field should be set to "No"
      */
     if (
-      formData?.["defineInitiative_projectedEndDate"]?.[0].value === "No" &&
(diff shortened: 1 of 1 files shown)


britt-mo (approved):
tested locally, didn't error out after tabbing through.

ailZhou (approved):
Tested in browser, did not get an error when tabbing and trying to save

PR #286, opened 2023-12-12
Make funding source and objectives modals match view only experience

gmrabian:
### Description

As per design we want the view only experience for these modals to have the "Save" button disabled, rather than changing the text and enabling it.

### Related ticket(s)

CMDCT-3132

---
### How to test

- Sign in as a state user
- Create a WP
- Create an initiative
- In the initiative, create an objective and a funding source
- Logout
- Sign in as an admin
- View the initiative you created and verify the objective and funding source modals have a "Save" button that is disabled

---
### Author checklist

---


britt-mo (approved):
tested locally, `Save` button is disabled w/ text unchanged

ailZhou (approved):
tested in deploy, looks good!

PR #294, opened 2023-12-13
[MFP] SAR - Dashboard alert banner and Add new SAR submission button

ajaitasaini:
### Description

This PR takes care of two tickets, making sure that the dashboard alert banner shows up and the add new SAR submission button is disabled when expected. We did this by specifically only checking for a Work Plan whose status is "Approved" and using that to determine whether we can add a SAR.

### Related ticket(s)

CMDCT-3016
CMDCT-187

---
### How to test

**NOTE: _I would recommend pulling down locally_**
1. Login. Before creating any WP, go to SAR and confirm you can see the alert banner. The add new SAR button should be disabled. 
2. Create a WP. Before getting it approved, go to SAR and confirm you can see the alert banner. The add new SAR button should be disabled. 
3. Login as admin and approve the work plan.
4. Log back in as state user, go to SAR. You should no longer see the alert banner and the button should be enabled. 

### Important updates

---
### Author checklist

---


gmrabian (approved):
Looks good! Worked as described. I did notice the error banner briefly pops up and then disappears once (I assume) we load the WPs in the background. Not sure if it'd be better to show banner after load instead of before this is minor anyway

https://github.com/Enterprise-CMCS/macpro-mdct-mfp/assets/57802560/307ba2df-e303-4749-9dad-f6cb3eb334e6

PR #302, opened 2023-12-14
Remove styling from html where it did nothing and to avoid console error

gmrabian:
### Description

React Fragments cannot accept properties other than key and children.

Because of how we do content parsing any blob provided as `"type": "html"` is rendered inside a react fragment
[html assignment](https://github.com/Enterprise-CMCS/macpro-mdct-mfp/blob/main/services/ui-src/src/utils/other/parsing.ts#L30)
[element creation logic](https://github.com/Enterprise-CMCS/macpro-mdct-mfp/blob/main/services/ui-src/src/utils/other/parsing.ts#L62-L68)

This was causing a console error when "style" was passed into the fragments properties. It turns out only one piece of content of type html was using props, and it didn't actually change the view, so I removed it. Now the error doesn't show up

Feedback:
- Do we want to instead remove styling from anything that is of type html? this feels wasteful because then you might not know that while writing content, and it also creates room for unnecessary code where you may have style props on an html type and it's being discarded on render anyway
- Is there a good place where we can note not to use properties in the json when working with type html?

### Related ticket(s)

CMDCT-3110

---
### How to test

- Log in as a state user
- Create a WP
- navigate to initative instructions in the report
- Verify the spacing is not affected by this change (compare to mdct dev)
- Verify no console errors on that page (refresh to verify render)

Video comparison of branch and dev:
https://github.com/Enterprise-CMCS/macpro-mdct-mfp/assets/57802560/e706b310-5ab7-4b91-973e-727a36d820c1

---
### Author checklist

---

### DIFF
--- a/services/app-api/forms/wp.json
+++ b/services/app-api/forms/wp.json
@@ -625,12 +625,7 @@
                 },
                 {
                   "type": "html",
-                  "content": "<br>If a recipient updates the MFP Work Plan to indicate that an initiative will no longer be sustained with MFP funding or state or territory-equivalent funding, the recipient must explain whether the initiative will be terminated or sustained through another funding source.<br><br>",
-                  "props": {
-                    "style": {
-                      "paddingBottom": "2rem"
-                    }
-                  }
+                  "content": "<br>If a recipient updates the MFP Work Plan to indicate that an initiative will no longer be sustained with MFP funding or state or territory-equivalent funding, the recipient must explain whether the initiative will be terminated or sustained through another funding source.<br><br>"
                 },
                 {
                   "type": "text",


karla-vm (approved):
re: your questions on the PR - I think this is one of those places where the `style` isn't necessary, but I think we do have a couple of instances where it has been. I'd rather style on the components if possible, but not sure how we go about those cases.

britt-mo (approved):
I just looked through the sar and wp json and didn't see the `style` prop used directly under `html` , only like so:
```
{
                  "type": "html",
                  "content": ""
                },
                {
                  "type": "ol",
                  "content": "",
                  "props": {
                    "styleType": "upper-roman",
                    "style": {
                      "marginLeft": "1.5rem",
                      "padding": "1rem"
                    }
                  },
```

So I feel like it's fine to move forward with the intention of avoiding using `style` within an `html` type, and unless I missed some areas, I don't think we need to remove it anywhere. As far as documenting this as best practice, I'm not sure :(

gmrabian:
Yeah I'm not sure where to document it either :/ 

Thanks for the reviews! I'm hoping it'll be obvious in the future because 1. it doesn't apply the styling and 2. there are console errors 🤷

braxex:
@gmrabian @karla-vm @britt-mo 

Good conversation here. A couple notes:

1. You're correct that the `html` type cannot accept styles in that way. 
2. The preferred method for doing styles for those html strings is to write them in JSS on the page. You can use standard CSS selectors to target the elements, including using `.some-class` in the JSS and inserting a class into the html string, like `<p class="some-class">blah blah blah</p>`. _Note: Doing all the styling on the page so it's all in one place and not spread through the json is preferable as well so that the json gets touched as little as possible and we make as few different versions of the json as possible. This includes things like using styling for direct control of spacing instead of using `<br>` breaks in the html string, which can be less reliable and has frustrated design in the past._
3. Notwithstanding the note under 2 above, you can also use inline styles inside those html strings (but prob don't) like `<p style="border:1px solid red;">woooo</p>`.

PR #308, opened 2023-12-18
Cmdct 3141 - prevent return key from exiting the modal and navigating back to prev page

britt-mo:
### Description
Hitting `return` in the overlay modal type exited the modal and navigated back to the previous page.

https://github.com/Enterprise-CMCS/macpro-mdct-mfp/assets/22454493/8000f15f-f962-4d4a-ace8-f3f236fa08dc

 I changed the button `type` from `submit` to `button`, and the `return` key tabs to the next field instead of exiting the modal to fix this issue.

### Related ticket(s)

CMDCT-3141
---
### How to test
In `main`, go to the state or territory initiatives and try to add an evaluation plan or funding source--fill out one or two inputs and then hit `return`, see that you've been sent back to the previous page. 
In project branch do the same thing and see that the `return` acts like a `tab`, going to the next input section instead of exiting the modal (unless all required fields have been filled out, in which case the `return` will hit the `submit` button.

### Important updates

---
### Author checklist

---


gmrabian (approved):
Works! Just to be clear: it still closes the modal if you hit `return` before you select a field. if you are in a field (other than textbox) it'll move on as expected

britt-mo:
> Works! Just to be clear: it still closes the modal if you hit `return` before you select a field. if you are in a field (other than textbox) it'll move on as expected

@gmrabian yeah, it closes the modal before you select a field, that's the same behavior in MCR and I think the selected/active element when you first open a modal is gonna be the `x Close` button. I think that behavior is fine, since you end up on the same page, vs getting sent back one page.

cassandradanger (approved):
looks good to me! totally cool with it closing if you hit enter right away, because the "close" button is default selected

PR #311, opened 2023-12-18
[CMDCT-3145] Report period and year fix

ntsummers1:
### Description

https://github.com/Enterprise-CMCS/macpro-mdct-mfp/assets/19439679/03f8681f-7f42-44d1-b7d9-65edf14ea622

### Related ticket(s)

CMDCT-3145

---
### How to test

Make sure that your isCopyOverTest is set to true on line 44 in AddEditReportModal `const isCopyOverTest = true`
Create a work plan and submit it as a state user
Approve that work plan as an admin user
Copy over the work plan as a state user
See that its been copied into the new year!

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/app-api/handlers/reports/create.ts
+++ b/services/app-api/handlers/reports/create.ts
@@ -143,7 +143,7 @@ export const createReport = handler(
       unvalidatedMetadata?.copyReport?.isCopyOverTest;
 
     if (overrideCopyOver) {
-      reportYear = reportPeriod == 2 ? reportYear++ : reportYear;
+      reportYear = reportPeriod == 2 ? reportYear + 1 : reportYear;
       reportPeriod = reportPeriod == 1 ? 2 : 1;
     }
 
@@ -278,7 +278,7 @@ export const createReport = handler(
         versionNumber: formTemplateVersion?.versionNumber,
         submissionName: createReportName(
           reportTypeExpanded,
-          currentDate,
+          reportPeriod,
           state,
           reportYear,
           workPlanMetadata
--- a/services/app-api/handlers/reports/create.ts
+++ b/services/app-api/handlers/reports/create.ts
@@ -41,7 +41,7 @@ import {
 import { getOrCreateFormTemplate } from "../../utils/formTemplates/formTemplates";
 import { logger } from "../../utils/logging";
 import { APIGatewayProxyEvent } from "aws-lambda";
-import { copyFieldDataFromSource, overriderDate } from "../../utils/other/copy";
+import { copyFieldDataFromSource } from "../../utils/other/copy";
 
 export const createReport = handler(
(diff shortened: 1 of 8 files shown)


services/app-api/utils/constants/constants.ts
```diff
@@ -16,6 +16,8 @@ export const error = {
   NO_TEMPLATE_NAME: "Must request template for download.",
   INVALID_TEMPLATE_NAME: "Requested template does not exist or does not match.",
   NOT_IN_DATABASE: "Record not found in database.",
+  UNABLE_TO_COPY:
+    "Unable to copy over report if todays date is the same period and year as previous report.",
```
cassandradanger, line 20: maybe we want to pass this by product? My only update would be
"Unable to copy over report if todays date is `in` the same period and year as a previous report."

cassandradanger, line 20: maybe we want to pass this by product? My only update would be
"Unable to copy over report if todays date is in the same period and year as a previous report."

cassandradanger (dismissed):
i think this looks good!! thanks for the quick turnaround!
one little nit - but not holding up approval for it - will re-approve if needed :)

services/app-api/utils/constants/constants.ts
ntsummers1: This is a message that only gets sent back as an api error to help us debug, so its really only help to us for debugging purposes. That said, I can make the change!

ntsummers1, line 20:
```suggestion
    "Unable to copy over report if todays date is in the same period and year as a previous report.",
```

services/app-api/handlers/reports/create.ts
```diff
@@ -135,14 +127,35 @@ export const createReport = handler(
...
+        ? calculatePeriod(currentDate, workPlanMetadata)
+        : workPlanMetadata!.reportPeriod;
+
+    const overrideCopyOver =
+      unvalidatedMetadata?.copyReport &&
+      unvalidatedMetadata?.copyReport?.isCopyOverTest;
+
+    if (overrideCopyOver) {
+      reportYear = reportPeriod == 2 ? reportYear + 1 : reportYear;
+      reportPeriod = reportPeriod == 1 ? 2 : 1;
```
britt-mo, line 147: I'm having trouble understanding line 147.  If it's an override copy over, and the report period is `1`, then it should be `2`, otherwise it'll be set to `1`?

gmrabian (commented):
Code changes look reasonable to me. Could you explain what was happening with dev and val and why the year was incorrectly incremented in val? I'm also curious how copy over plays into that, because I tested it by archiving existing reports and creating a new, non-copied one

gmrabian (approved):

britt-mo (approved):

PR #314, opened 2023-12-20
Update PDF table per design requests

gmrabian:
### Description

The Transition Benchmarks PDF table had some more rules to follow. Some of these didn't work due to "Answered" being capitalized instead of lowercase in some equality checks. See How to test for more details on the rules we are trying to follow

### Related ticket(s)

CMDCT-3151

---
### How to test

- Log in as a state user
- Create a WP
- Open the PDF and view the `Transition Benchmark Projections` table
- Answer the questions on the transition benchmarks page and verify the table follows these rules as you do so (you'll need to refresh the pdf page after making changes):
  • If a user doesn't answer the drawer (radio buttons) for Transition Benchmarks Projections for a target population (it's still in its default state), then show "Not answered" on the PDF table column for that population (in color-error-darker red). Any sums of cells that include at least 1 "Not answered" should have an asterisk in the total column or row
  • If a user answers "No" in the drawer, put N/As down the entire column for that population in the PDF
  • If a user answers "Yes" in the drawer, the answer can contain a number, 0, or "N/A" and it all counts as answered (should not lead to an asterisk in the total column unless a cell in its column or row has "Not answsered" for another population.

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/components/export/ExportedModalDrawerReportSection.tsx
+++ b/services/ui-src/src/components/export/ExportedModalDrawerReportSection.tsx
@@ -44,8 +44,8 @@ export const ExportedModalDrawerReportSection = ({
     let quarterArray = [];
 
     if (
-      entity?.transitionBenchmarks_applicableToMfpDemonstration?.[0].value !==
-      "No"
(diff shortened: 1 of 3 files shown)


ajaitasaini (approved):
woohoo!! good job

PR #319, opened 2023-12-21
SAR - Resubmission Question

keeysnc:
### Description
Added resubmission field in SAR. The question field is visible:
1. If report is SAR
2. If a report is in  'In Revision'
3. If a report has been submitted more than 1 time and is being re-revised

### Related ticket(s)
https://jiraent.cms.gov/browse/CMDCT-3015

https://github.com/Enterprise-CMCS/macpro-mdct-mfp/assets/14514294/e9b898ae-80b1-4ea0-a5a8-13b338f1d926

---
### How to test
1. Create a WP as a state user and submit it

2. Approve WP report as an admin

3. Create a Sar with state user ( field should be hidden in general information)

4. Submit SAR

5. Log in as admin and unlock the report ( field should be visible after unlocking)

6. Log back in as state user and you ( field should be visible and cleared of any value)

7. Fill out the field and resubmit the SAR 

8. Log back in as an admin and the field should display new revision text in field

9. If you unlock the SAR again as an admin and return to the SAR as a state user the 'resubmission' field should reset to an empty field

---
### Author checklist

---


karla-vm (approved):
LGTM, awesome work on this one, it's a tricky use-case! 🥇

gmrabian (approved):

gmrabian (commented):
If you submit as a state user, before unlocking as an admin, the revision question shows up with "N/A" if you go back in to view the report. Is this ok? Seems like it should only show up after it's been unlocked the first time, though maybe not a big deal. I'm just imagining a state user might submit and then go to print the report and a question about resubmission shows up at the top that they aren't expecting

keeysnc:
> If you submit as a state user, before unlocking as an admin, the revision question shows up with "N/A" if you go back in to view the report. Is this ok? Seems like it should only show up after it's been unlocked the first time, though maybe not a big deal. I'm just imagining a state user might submit and then go to print the report and a question about resubmission shows up at the top that they aren't expecting

Hm, I can take a quick look at this.

keeysnc:
> If you submit as a state user, before unlocking as an admin, the revision question shows up with "N/A" if you go back in to view the report. Is this ok? Seems like it should only show up after it's been unlocked the first time, though maybe not a big deal. I'm just imagining a state user might submit and then go to print the report and a question about resubmission shows up at the top that they aren't expecting

Updated.

karla-vm (approved):

services/ui-src/src/components/reports/StandardReportPage.tsx
```diff
@@ -58,6 +59,42 @@ export const StandardReportPage = ({ route, validateOnRender }: Props) => {
...
+    ) {
+      return formDataCopy;
+    } else {
+      const removeResubmissionDataFields: any[] = formDataFields.filter(
+        (data: AnyObject) =>
+          data.id !== "generalInformation_resubmissionInformation"
+      );
+      formDataCopy.fields = [...removeResubmissionDataFields];
+
+      return formDataCopy;
```
gmrabian, line 94: the `else` and the first `if` now have the same logic, right?

gmrabian, line 94: the else and the first if now have the same logic, right?

services/ui-src/src/components/reports/StandardReportPage.tsx
```diff
@@ -58,6 +59,42 @@ export const StandardReportPage = ({ route, validateOnRender }: Props) => {
     navigate(nextRoute);
   };
 
+  const filterResubmissionData = (formData: FormJson) => {
+    //make deep copies of formData
+    const formDataCopy = JSON.parse(JSON.stringify(formData));
+    const formDataFields = JSON.parse(JSON.stringify(formData?.fields));
```
gmrabian, line 65: suggestion: there's a native js function for this now called `structuredClone()`! does basically the same thing as json.parse(json.stringify)

gmrabian (commented):
Form logic works as expected! Can simplify the code a little I think

services/ui-src/src/components/reports/StandardReportPage.tsx
keeysnc: Same logic but different condition. I could put it in a function..

services/ui-src/src/components/reports/StandardReportPage.tsx
keeysnc: oh nice! Thanks

gmrabian (approved):

gmrabian (changes requested):
re: tests failing on structuredClone - oh yeah older versions of jest don't have this yet :/ we encountered this elsewhere in MFP.

If you add this to the test file it will fix it. We do that elsewhere in MFP
```
global.structuredClone = jest.fn((val) => {
  return JSON.parse(JSON.stringify(val));
});
```

gmrabian (approved):
I think we got there 💪

karla-vm (approved):

PR #340, opened 2024-01-02
CMDCT-3149: Enable Reset Work Plan Functionality

karla-vm:
### Description

Allowing users to reset a "continued" Work Plan for their next reporting period.

**Note:** I also fixed a couple of verbiage issues on the `wp.json`

### Related ticket(s)

CMDCT-3149

---
### How to test

- Log in as a state user
- Create and submit a new Work Plan
- Log in as an admin
- Approve that Work Plan
- Log BACK in as the same state user
- Click on the `Continue Work Plan for next Period` button
- On the modal, click on `Reset MFP Work Plan`

Screenshot 2024-01-02 at 2 15 49 PM

You should be able to see, fill, and submit a fresh WP on the dashboard ✨ 

### Important updates

N/A

---
### Author checklist

---

### DIFF
--- a/services/app-api/forms/wp.json
+++ b/services/app-api/forms/wp.json
@@ -42,6 +42,10 @@
                 "target": "_blank",
                 "aria-label": "User Guide and Help File (Link opens in new tab)"
               }
+            },
+            {
+              "type": "html",
+              "content": "."
             }
           ]
         },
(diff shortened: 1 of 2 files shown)


gmrabian (approved):
Looks good! 
Two comments:
1. This is a silly question but since it creates a new report in the same reporting period I'm assuming it's ok to have two reports with totally different info for the same period?
2. The ticket says it should start as "In Progress" but it shows up "Not started". I don't think this is an issue, but maybe worth checking with product or design if this is ok.

karla-vm:
> Looks good! Two comments:
> 
> 1. This is a silly question but since it creates a new report in the same reporting period I'm assuming it's ok to have two reports with totally different info for the same period?
> 2. The ticket says it should start as "In Progress" but it shows up "Not started". I don't think this is an issue, but maybe worth checking with product or design if this is ok.

I will be checking with design on both of these! Great questions, Garrett, thank you 🙏🏽

PR #362, opened 2024-01-04
Fix formtemplate and fieldData getting saved in WP metadata upon SAR creation

gmrabian:
### Description

When creating a SAR we also update the associated WP. We fetch the WP, [spread its metadata](https://github.com/Enterprise-CMCS/macpro-mdct-mfp/blob/main/services/app-api/handlers/reports/create.ts#L308), adding the `associatedSar` field, and update it. The problem is this uses our report fetch, which [returns metadata, formTemplate, and fieldData all together](https://github.com/Enterprise-CMCS/macpro-mdct-mfp/blob/main/services/app-api/handlers/reports/fetch.ts#L98-L104) for the frontend. Then `formTemplate` and `fieldData` mistakenly get spread in as metadata when we add the field for the SAR.

This takes a structured clone of the fetch and removes those two fields so that the metadata is pure.

### Related ticket(s)

CMDCT-3167

---
### How to test

- As a state user, create and submit a WP
- As an admin user, approve that WP
- Open up dynamodbadmin and check the size of the WP metadata
- As a state user, create a SAR
- Refresh dynamodbadmin and ensure the WP metadata only added one line for `associatedSar`

---
### Author checklist

---

### DIFF
--- a/services/app-api/utils/other/other.ts
+++ b/services/app-api/utils/other/other.ts
@@ -91,9 +91,12 @@ export const getLastCreatedWorkPlan = async (
     const workPlan = await fetchReport(fetchWPReportEvent, _context);
     const workPlanBody: ReportShape = JSON.parse(workPlan.body);
 
-    // And get its metadata and field data from the response and return it!
-    const workPlanMetadata: ReportMetadataShape = workPlanBody;
+    // fetchReport returns fieldData and formTemplate together with metadata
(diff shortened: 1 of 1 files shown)


cassandradanger (approved):
tested the WP and SAR in the browser and it looks like everything is good! TY

ailZhou (approved):
tested locally and not see the wp data get copied again. thank you for catching this data error!

PR #372, opened 2024-01-09
Add tests for getEntityStatus

benmartin-coforma:
### Description
Some new tests, with two changes to "real" code:
1. Fix the type for EntityRow.formEntity.
   * This was set to the wrong type (ReportJson), but it was being cast to `any` on the way in (in EntityDetailsDashboardOverlay) and on the way out (to getEntityStatus.getInitiativeDashboardStatus)
   * This is almost a no-op change; Typescript type annotations have no effect on runtime behavior.
   * However it _did_ expose a bug with the `EntityRow.key`; I believe all rows were getting the same key. Now they will get an index-based key.
2. Fix a potential bug when detecting optional fields in getCloseoutStatus
   * Sometimes validation is a string, sometimes it's an object.
   * I don't believe this bug ever manifested; if I recall correctly, all of our optional fields currently have validation as a string in the form template.
   * But it's better to be safe than sorry! Maybe we'll have such fields in the future.

### Related ticket(s)
CMDCT-3121

---
### How to test
`yarn test --coverage`

### Important updates
n/a

---
### Author checklist

### DIFF
--- a/services/ui-src/src/components/overlays/EntityDetailsDashboardOverlay.tsx
+++ b/services/ui-src/src/components/overlays/EntityDetailsDashboardOverlay.tsx
@@ -106,9 +106,9 @@ export const EntityDetailsDashboardOverlay = ({
             initiativeName={selectedEntity?.initiative_name}
           />
           <Table content={tableHeaders()}>
-            {entitySteps?.map((step: any) => (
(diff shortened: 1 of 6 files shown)


cassandradanger (approved):
lgtm!

cassandradanger (approved):

PR #379, opened 2024-01-11
CMDCT-3181: [Bug] WP - Define Initiative

karla-vm:
### Description

Currently, in the `WP Define Initiative` step, if a user selects `No` to the last question of the form, they are not able to submit using the `Save & return` button. 

Screenshot 2024-01-11 at 1 54 36 PM

The issue was that the "Yes" option has a nested child input field, but it was not marked as `nested` in the `wp.json` and users weren't able to submit because the form was "incomplete" unless that field was filled.

### Related ticket(s)

CMDCT-3181

---
### How to test

You should be able to select `No` and submit this form now.

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/app-api/forms/wp.json
+++ b/services/app-api/forms/wp.json
@@ -945,7 +945,8 @@
                               "validation": {
                                 "type": "endDate",
                                 "dependentFieldName": "defineInitiative_projectedStartDate",
-                                "parentFieldName": "defineInitiative_projectedEndDate"
+                                "parentFieldName": "defineInitiative_projectedEndDate",
+                                "nested": true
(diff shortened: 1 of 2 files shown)


karla-vm:
> thanks for fixing this karla! does this change not break end date validation? there was a bug i worked on earlier to add that validation in, and i believe the validation wasn't being set due to the logic we have around nested fields: #364

ah, you're absolutely right! I need to do some more digging. thanks for pointing this out!

britt-mo (approved):

cassandradanger (approved):
lgtm!

PR #381, opened 2024-01-16
[CMDCT-3175] The Initiatives Versioning Experience

ntsummers1:
### Description
This PR adds a new type of form template transformation, which can generate entire new forms. It is used only in the SAR, for the new pageType `dynamicModalOverlay`.

We already had route-level transformations, which can remove an entire route from the form template if appropriate. We also had field-level transformations, which can adjust the content of a single field, or create clones of a field (for example, to create a field for each target population). This transformation takes a template containing entity steps, and duplicates that template for each initiative. That lets us fill each entity step's form with questions that are specific to that initiative.

The most notable changes are:
1. API: `sar.json` - The `State- or Territory-Specific Initiatives` route, once a `modalOverlay`, is now a `dynamicModalOverlay`. The entitySteps (and some related data) are now wrapped in a `template` element.
2. API: `transformations.ts` - All of the form template transformation code previously in formTemplates.ts has been moved to here. It also contains new code to duplicate the `template` for each initiative, in `generateSARFormsForInitiatives()`. The clones will live in the `initiatives` array in the form template.
3. API: `formTemplates.ts` - Besides the transformation code being moved out, `compileValidationJsonFromRoutes` has also been changed. It now knows how to find fields within `initiatives` (which will have been populated with initiative-specific data by the time this function runs).
4. API: `completionStatus.ts` - Now knows how to find entity steps within `initiatives`.
5. UI: `DynamicModalOverlayReportPage.tsx` - Imitates the existing `ModalOverlayReportPage`, but knows how to find forms within `initiatives`.

There are 20 other changed files, but their changes can be categorized as
* Adjustments to type definitions
* Accounting for the new form template structure
* Minor bug fixes
* And a wee bit of dead code removal. Goodbye, `overlayForm`, we didn't need ye.

Finally, note that this PR does not include full functionality for completionStatus.ts in the API. It will sometimes say that a report is complete, when one of its initiatives is actually not complete. This work is ongoing.

### Related ticket(s)
Resolves:
* CMDCT-3175
* CMDCT-3176

Related to:
* CMDCT-44
* CMDCT-36

---
### How to test
* Create a Work Plan
  * Include multiple initiatives, with multiple funding sources
* Create a SAR from that work plan
* Navigate to the state- and territory-specific initiatives
* Note that each initiative is represented in the list
* Click into each initiative
  * Note that relevant information related to objectives progress has been autofilled into read-only fields
  * Note that each initiative has questions that are specific to its funding sources
* Note that field data can be saved as expected
* Note that sections are marked complete when completed
* Note that UI generally resembles the design 😉 

### Important updates
This change is ⚠️ NOT backward-compatible ⚠️ with existing SARs. They will all have to be deleted and re-created.

Recreating an existing SAR is also non-trivial; each Work Plan remembers whether it has been used to create a SAR, and refuses to be used to create a second one. That WP data would need to be changed in AWS.

Hopefully there is no need to do this. If we have any existing SARs in production or val, which weren't created by MDCT team members for testing purposes, then that conversion will need attention from MDCT devs, and possible a migration.

---
### Author checklist


services/app-api/tsconfig.json
benmartin-coforma, line 1: These changes are in accordance with https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping

It would be nice, in the future, to replace our use of `date-fns-tz` with `Intl.DateTimeFormat`. For one, because the latter is built into the language. For two, because that will allow us to upgrade our module resolution strategy. `node` is considered an old strategy that exists for backwards compatibility. https://www.typescriptlang.org/tsconfig#moduleResolution

services/ui-src/src/components/overlays/EntityDetailsDashboardOverlay.tsx
```diff
@@ -32,7 +33,18 @@ export const EntityDetailsDashboardOverlay = ({
...
+    | undefined;
+  if (route.pageType === "dynamicModalOverlay") {
+    entitySteps = (
+      route as DynamicModalOverlayReportPageShape
+    ).initiatives.find(
+      (initiative) => initiative.initiativeId === selectedEntity?.id
+    )?.entitySteps;
+  } else {
+    entitySteps = route.entitySteps;
+  }
```
benmartin-coforma, line 47: I don't like this, and I intend to change it before this PR is merged.

services/ui-src/src/components/reports/DynamicModalOverlayReportPage.tsx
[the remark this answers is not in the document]
benmartin-coforma: Actually I was able to simplify the code considerably without any new data structures.

cassandradanger (commented):
only thing I found in the browser experience is if a funding source is "other", i'm guessing we want to display what text was entered in the WP input. (i'm verifying with Kim)
I'm also okay if this issue is addressed in a separate PR

UPDATE FROM KIM - this should display the user's input - [figma has been updated](https://www.figma.com/file/oOJMUZerbafZefLQQ33GwI/MFP_Playground?type=design&node-id=7090%3A100375&mode=design&t=CHXfod4JO6VZGUjm-1)
Screenshot 2024-01-22 at 4 50 54 PM

cassandradanger (approved):

ailZhou (approved):
having been using this in another branch and so far only one bug was found but it got fixed

PR #388, opened 2024-01-19
SAR - Objectives progress- State or Territory Specific Initiatives

ajaitasaini:
### Description

This PR covers the Objectives progress modal and making sure quantitative quarters show, if applicable.

NOTE: as of right now, the objectives modal will not include the PDF-related hint text. it will be addressed in a future PR, along with adding back in the Review PDF button + hint text at the bottom of the Objectives Progress step.

### Related ticket(s)

CMDCT-44

---
### How to test

1. Login, fill out WP, ensuring that at least one objective in an evaluation plan includes quantitative targets (check Yes here to see the dropdown, and fill with values
Screenshot 2024-01-23 at 6 20 49 PM

2. Submit work plan, approve as admin, and login to SAR 
3. Navigate to State- or Territory-Specific Initiatives and navigate to Initiatative > Objectives > Objective card with the quantitative quarters and confirm that you're seeing this UI 

image

4. Confirm that an objective card that does not have quantitative targets still has this UI 
image

### Important updates

---
### Author checklist

---


benmartin-coforma (dismissed):
This looks great, well done Ajaita and Aileen.

services/ui-src/src/components/reports/OverlayModalPage.tsx
```diff
@@ -26,14 +26,20 @@ export const OverlayModalPage = ({
   closeEntityDetailsOverlay,
   route,
 }: Props) => {
-  const { verbiage, modalForm, stepType } = route;
+  const { verbiage, stepType } = route;
   const { report, selectedEntity, setSelectedEntity, editable } = useStore();
   const [selectedStepEntity, setSelectedStepEntity] = useState<
     EntityShape | undefined
   >(undefined);
+  const [selectedEntityIndex, setSelectedEntityIndex] = useState<number>(0);
```
ntsummers1, line 34: All of this code looks pretty good, and runs well too! I think this is the one area I'd like to see an update on. I don't think we need this state object. We have the selectedStepEntity which knows the id thats been selected, and the cards have an objective id that matches from the transformation. We can use that to match the form up.

A function could look something like this: 
```
  const getModalForm = () => {
    if (objectiveCards) {
      for (let card of objectiveCards) {
        if (card.modalForm.objectiveId === selectedStepEntity?.id)
          return card.modalForm;
      }
    } else if (route?.modalForm) {
      return route.modalForm;
    }
    return { id: randomUUID(), formFields: [] };
  };

  let modalForm = getModalForm();
```

That last return is important too - If, for some reason, there isn't an objective card the route won't have a modalForm and thus the page would error out. I don't forsee this happening but it would be good to cover our bases.

services/ui-src/src/components/reports/OverlayModalPage.tsx
```diff
@@ -50,6 +56,12 @@ export const OverlayModalPage = ({
     }
   }, [report]);
 
+  useEffect(() => {
+    //only update if we're using objectiveCard Templates
+    if (objectiveCards)
+      modalForm = objectiveCards[selectedEntityIndex].modalForm;
+  }, [selectedEntityIndex]);
+
```
ntsummers1, line 64: If you went with the above comment you can swap this to 
```
  useEffect(() => {
    getModalForm();
  }, [selectedStepEntity]);
```

services/ui-src/src/components/reports/OverlayModalPage.tsx
```diff
@@ -66,8 +78,9 @@ export const OverlayModalPage = ({
     onClose: addEditEntityModalOnCloseHandler,
   } = useDisclosure();
 
-  const openAddEditEntityModal = (entity?: EntityShape) => {
+  const openAddEditEntityModal = (entity?: EntityShape, index?: number) => {
     if (entity) setSelectedStepEntity(entity);
+    setSelectedEntityIndex(index!);
```
ntsummers1, line 83: And would be able to remove this line and the index param

services/app-api/utils/transformations/transformations.ts
```diff
@@ -376,6 +392,81 @@ export const fundingSources = (
...
+
+  const initiativeToUse = workPlanFieldData?.initiative.find(
+    (initiative: any) => initiative.id === initiativeId
+  );
+
+  const objectiveToUse = initiativeToUse.evaluationPlan.find(
+    (objective: any) => objective.id === objectiveId
+  );
+
+  delete fieldToRepeat.transformation;
```
ntsummers1, line 413:
```suggestion
  delete fieldToRepeat.transformation;
  delete fieldToRepeat.objectiveCardTemplate;
```

ntsummers1: I think you want to delete this template too

services/ui-src/src/components/reports/OverlayModalPage.tsx
ntsummers1: Could even get rid of that last return since our Typescript is saying a modalForm is guaranteed to exist on the route. So maybe a modalForm should live in the sarJson if nothing but to serve as a empty id and formField to be overwritten.

That would mean the above code looks more like this:
```
  const getModalForm = () => {
    if (objectiveCards) {
      for (let card of objectiveCards) {
        if (card.modalForm.objectiveId === selectedStepEntity?.id)
          return card.modalForm;
      }
    } 
    return route.modalForm;
  };

  let modalForm = getModalForm();
```

ntsummers1 (changes requested):

services/ui-src/src/components/reports/OverlayModalPage.tsx
ailZhou: Great catch, thanks for breaking it down! suggestion has been implemented with a slight modification for selectedStepEntity being undefined at the start,

services/ui-src/src/components/reports/OverlayModalPage.tsx
ailZhou: this useEffect ended up not being needed once i implemented the function above, so its been removed

services/ui-src/src/components/reports/OverlayModalPage.tsx
ailZhou: removed and lost in the ether

ntsummers1 (approved):
Thanks for making those!

benmartin-coforma (approved):

PR #434, opened 2024-02-05
Restrict requests by assigned state for state users

gmrabian:
### Description

Copies work from [MCR](https://github.com/Enterprise-CMCS/macpro-mdct-mcr/pull/11575)

We have an auth utility called `hasPermissions` that verifies the requesting user has the role expected for a given request (ex: ADMIN for create banner, STATE USER for submit report, etc.)

Key changes:
- `services/app-api/utils/auth/authorization.ts` the `hasPermissions()` method is updated to optionally accept `state`. It performs different verification logic as a result.
  - also added the `isAuthorizedToFetchState()` method to verify that if they are a state user, they are a state user for the requested state. Admins and other types can request any state.

### Related ticket(s)

CMDCT-3197

---
### How to test

To verify the `create` updates (same logic as `update` and `submit`)
- Log in as a state user
- With the network tab open, create a WP report
- Manipulate the POST network request to send it to a different state
- Verify that request returns a 403 (on `main` it is a 201 😬 )

You can do a similar test to verify `fetch` permissions checks
- Log in as a state user
- With the network tab open, refresh the WP reports dashboard
- Manipulate the GET network request to request reports from a different state
- Verify that request returns a 403 (on `main` it is a 200 😬 )

You can also verify admin functions still work as expected (like create and delete banner)

On Chrome you can right-click the command and select "Copy as cURL or Copy as Fetch". The former you can paste in a terminal and edit. The latter you can paste in the browser's console and edit (I found this easier). Resend the command and verify the status code.

---
### Author checklist

---

### DIFF
--- a/services/app-api/handlers/reports/create.ts
+++ b/services/app-api/handlers/reports/create.ts
@@ -46,16 +46,9 @@ import { extractWorkPlanData } from "../../utils/transformations/transformations
 
 export const createReport = handler(
   async (event: APIGatewayProxyEvent, _context) => {
-    if (!hasPermissions(event, [UserRoles.STATE_USER])) {
-      return {
-        status: StatusCodes.UNAUTHORIZED,
-        body: error.UNAUTHORIZED,
-      };
-    }
-
     const requiredParams = ["reportType", "state"];
 
-    // Return No_Key when not given a state and reportType as a paramater
+    // Return No_Key when not given a state and reportType as a parameter
     if (
       !event.pathParameters ||
       !hasReportPathParams(event.pathParameters, requiredParams)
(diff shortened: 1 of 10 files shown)


services/app-api/handlers/reports/create.ts
```diff
@@ -46,16 +46,9 @@ import { extractWorkPlanData } from "../../utils/transformations/transformations
...
-      return {
-        status: StatusCodes.UNAUTHORIZED,
-        body: error.UNAUTHORIZED,
-      };
-    }
-
     const requiredParams = ["reportType", "state"];
 
-    // Return No_Key when not given a state and reportType as a paramater
+    // Return No_Key when not given a state and reportType as a parameter
```
benmartin-coforma, line 51: Thank you, Garrett!

benmartin-coforma (approved):
I haven't tested this, but the code looks good!

karla-vm (approved):

PR #435, opened 2024-02-06
Add check for repeated work plan creation

gmrabian:
### Description

Currently our UI disabled a button when we don't want to allow the user to create multiple work plans. The API does not prevent this action, so you can circumvent the UI and create as many as you want 😱 

Open to feedback on status codes and error messaging for this situation!

### Related ticket(s)

CMDCT-3251

---
### How to test

- Using either your browser or the MFP Postman collection, attempt to create multiple Work Plans
  - In browser you can just try resending the network call multiple times
- Ensure the first one gets created. Any after that return 400.
  - If you complete a WP you can then create another. Then the pattern repeats.

---
### Author checklist

---

### Notes
I tried adding a test file: `other.test.ts` and it gave many errors when I only tried to test the name formatting function. Maybe I need to mock the fetch calls even though they weren't in my test? Idk but if coverage is low I may need someone to look at that with me.

### DIFF
--- a/services/app-api/handlers/reports/create.test.ts
+++ b/services/app-api/handlers/reports/create.test.ts
@@ -115,6 +115,9 @@ describe("Test createReport API method", () => {
   beforeEach(() => {
     jest.clearAllMocks();
     jest.useRealTimers();
+    jest
+      .spyOn(helperFunctions, "eligibleToCreateNewWP")
+      .mockResolvedValueOnce(true);
   });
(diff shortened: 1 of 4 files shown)


berryd (approved):

ntsummers1 (approved):

karla-vm (approved):

benmartin-coforma:
I think the most semantically-appropriate HTTP status code would be 409 CONFLICT, but 400 does the job 👍

PR #436, opened 2024-02-06
[Bugfix] WP Copy Initiative Status in Review & Submit

ailZhou:
### Description

The strangest case of this should have been broken the whole time but somehow managed to not be until the copy. 

WP Review and Submit status for State- or Territory Specific Initiatives status would show only complete when a WP copy was made. The issue was that the status did not actually run because it was trying to check for an initiativeId that only existed in dynamic modal overlay report pages. This PR adds a simple check to the conditional.

### Related ticket(s)

CMDCT-3250

---
### How to test

- Sign into MFP, any state user
- Fill out and submit a WP
- Signout of state user and sign in as admin
- Approve the previously made WP
- Sign back into the state user, and copy the approve WP
- Go to into the copied WP, check Review and Submit.
- Check to make sure the status of State- or Territory Specific Initiatives is false
- Fill out State- or Territory Specific Initiatives and then check that the status is complete .

### Important updates

N/A

---
### Author checklist

---

### DIFF
--- a/services/app-api/utils/validation/completionStatus.ts
+++ b/services/app-api/utils/validation/completionStatus.ts
@@ -189,7 +189,11 @@ export const calculateCompletionStatus = async (
                 ? stepForm.form
                 : stepForm.modalForm;
 
-              if (nestedFormTemplate?.initiativeId !== stepFields?.id) {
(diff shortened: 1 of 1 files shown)


cassandradanger (approved):

gmrabian (approved):
works as described

PR #457, opened 2024-02-16
Fix nested field validation

benmartin-coforma:
### Description
In the field data, choice list option IDs start with the parent field ID, followed by the option ID. Therefore when we test to see if a choice is selected, we must match on the suffix of the option ID.

### Related ticket(s)
CMDCT-3277

---
### How to test
1. Create a WP with at least one applicable target population
2. Create a SAR based on that WP
3. Navigate to the RET disenrollment page
4. Confirm that patient counts disenrolled for "other" reasons must be numeric.

### Important updates
n/a

---
### Author checklist

### DIFF
--- a/services/ui-src/src/utils/validation/validation.test.ts
+++ b/services/ui-src/src/utils/validation/validation.test.ts
@@ -1,5 +1,8 @@
 import { mapValidationTypesToSchema } from "./validation";
 import * as schema from "./schemas";
+import { FormJson, isFieldElement } from "types";
+import { compileValidationJsonFromFields } from "utils/reports/reports";
+import * as yup from "yup";
 
 const mockStandardValidationType = {
   key: "text",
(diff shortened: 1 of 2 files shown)


gmrabian (approved):

britt-mo (dismissed):
works as expected! Thanks for figuring this out. Only failing cause of Code Climate

britt-mo (approved):

services/ui-src/src/utils/validation/validation.test.ts
```diff
@@ -76,3 +79,119 @@ describe("Test mapValidationTypesToSchema", () => {
...
+                  },
+                },
+              ],
+            },
+          ],
+        },
+      },
+    ],
+  };
+  let petSchema: any; // this type is heckin' complicated
```
gmrabian, line 122: 😆

gmrabian (approved):
mhmm

PR #472, opened 2024-02-23
Cmdct 3287 - hide resubmission content on PDF if SAR is not a resubmission

britt-mo:
### Description
Currently the Resubmission content always displays on the SAR PDF. This PR checks to see if the SAR is not a resubmission, and hides the content if so.

### Related ticket(s)

CMDCT-3287

---
### How to test
Create a SAR, check that the PDF does not contain `resubmission` content. Submit it, log in as an admin and unlock it, log in as a state user and add resubmission content, see that this content appears in the PDF now.

### Important updates

---
### Author checklist

---


services/ui-src/src/components/export/ExportedReportFieldTable.tsx
```diff
@@ -71,7 +72,17 @@ export const renderGeneralInformation = (
   pageType: string,
   entityType?: string
 ) => {
+  const { report } = useStore() ?? {};
+
+  // check if this was a resubmission
+  const isResubmission =
+    (report?.reportType === "SAR" && report?.submissionCount) ||
+    (report?.status === ReportStatus.SUBMITTED &&
+      report?.submissionCount! > 1 &&
```
gmrabian, line 81: with this current logic the resubmission question displays upon the first submit, before it's been released.

We have [similar logic](https://github.com/Enterprise-CMCS/macpro-mdct-mfp/blob/main/services/ui-src/src/components/reports/StandardReportPage.tsx#L78-L87) for this in the StandardReportPage component. Maybe it'd make sense to use the same logic so that the question showing up in the form is the same as when it shows up in the PDF?

gmrabian (changes requested):

keeysnc (approved):
lgtm!

keeysnc (approved):

keeysnc (approved):
Nice, test looks good now.

PR #480, opened 2024-03-01
Cmdct 3282 - update to generic error alert

britt-mo:
### Description
Design wanted the generic/global error alert message to be a bit more useful for the user. The agreed upon change was to tell the user to refresh and if the problem persists, to email the help desk.

Screenshot 2024-03-01 at 11 15 27 AM

If anyone has suggestions to change `link={"link"}`, let me know. Karla and I did that to avoid changing the `type`. Also, the styling can probably be cleaned up as well, but I was having a hard time with it.

### Related ticket(s)

CMDCT-3282

---
### How to test
Login as a state user, if you don't see the error banner when you first enter the workplan page, create a new workplan, enter it, turn your wifi off, and then leave the form. See that the error alert matches the spec. You can also enter the SAR page before having an approved work plan to see the placement of both the SAR error and generic global error.

Specifics to note:
alert should not have a "close" button per CMS Design System
alert should contain a clickable link to email `mdct_help@cms.hhs.gov` (this will not work in dev local)
alert should be same width as page content column
this alert should come above (to the top of) the alert that appears if a user does not have an approved Work Plan available for use in SAR

### Important updates

---
### Author checklist

---


ntsummers1 (dismissed):

services/ui-src/src/components/alerts/Alert.tsx
```diff
@@ -42,14 +46,16 @@ export const Alert = ({
...
+                    href={createEmailLink({ address: helpDeskEmail })}
+                    isExternal
+                  >
+                    {helpDeskEmail}
                   </Link>
-                </Text>
-              )}
+                )}
+                .
+              </Text>
```
braxex, line 58: Blocker: This is actually changing all instances of the Alert component to have this. You can see it here: 
Screenshot 2024-03-04 at 15 49 25

braxex: Blocker: If we want to implement real inline links here, we'll need to pass that info through the description's HTML parser on line 48. You could modify the verbiage file to something like this: 

```
const genericErrorContent = [
  {
    type: "span",
    content:
      "Something went wrong on our end. Refresh your screen and try again.<br/>If this persists, contact the MDCT Help Desk with questions or to request technical assistance by emailing ",
  },
  {
    type: "externalLink",
    content: "mdct_help@cms.hhs.gov",
    props: {
      href: "mailto:mdct_help@cms.hhs.gov",
      target: "_blank",
    },
  },
  {
    type: "span",
    content: ".",
  },
];
```

braxex (changes requested):
@karla-vm @britt-mo @ntsummers1 We'll need to make some changes here. I left some comments and made my own PR that implemented them, which you can see here: https://github.com/Enterprise-CMCS/macpro-mdct-mfp/pull/482. Hopefully all the notes make sense. Let me know if you'd like to chat about it.

Oh btw my PR is pointed at your branch so if you like it you'll have to merge to your branch first, then the trunk.

services/ui-src/src/verbiage/errors.ts
```diff
@@ -21,14 +21,14 @@ export const validationErrors = {
...
-  GET_REPORTS_BY_STATE_FAILED:
-    "Reports could not be loaded. Please contact support.",
-  SET_REPORT_FAILED: "Report could not be updated. Please contact support.",
-  DELETE_REPORT_FAILED: "Report could not be deleted. Please contact support.",
+  GET_REPORT_DATA_FAILED: genericErrorContent,
+  SET_REPORT_DATA_FAILED: genericErrorContent,
+  GET_REPORT_FAILED: genericErrorContent,
+  GET_REPORTS_BY_STATE_FAILED: genericErrorContent,
+  SET_REPORT_FAILED: genericErrorContent,
+  DELETE_REPORT_FAILED: genericErrorContent,
```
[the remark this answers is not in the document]
jessabean: +1 to not replacing these all with generic error. There is definitely room for improvement on error messaging here, but at the very least differentiating these error messages is helpful for developer experience and debugging!

services/ui-src/src/components/reports/ReportProvider.test.tsx
```diff
@@ -102,7 +107,7 @@ describe("ReportProvider", () => {
     const button = screen.getByText("Create");
     await userEvent.click(button);
     expect(screen.getByTestId("errorMessage")).toHaveTextContent(
-      /Report could not be updated/
+      /Something went wrong on our end. Refresh your screen and try again./
```
keeysnc, line 110: This text is in a lot of files. If we need to update this again it may be a little tedious. Could we mock this error message somewhere and import it into whatever tests need it?

keeysnc:
Blocker: From the figma mockup it seems like alert should be same width as page content column. 

Here's some additional styling based on the mockup as well.

ErrorAlert.tsx

`const sx = {
  root: {
    background:'palette.error_lightest',
    borderLeft: "palette.error",
    minHeight: 0,
    marginY: "1rem",
    maxWidth: "55.25rem",
    margin: "auto",
    borderInlineStartWidth: "0.5rem",
    borderInlineStartColor: 'palette.error',
    boxShadow: "0px 4px 3px rgba(0, 0, 0, 0.2)",
    "&.toast": {
      position: "absolute",
      right: 0,
      width: "90%",
      maxWidth: "500px",
      margin: "1rem",
      boxShadow: "0px 3px 9px rgba(0, 0, 0, 0.2)",
    },
  },
};`

services/ui-src/src/verbiage/errors.ts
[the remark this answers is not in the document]
jessabean: Just reiterating that I think we should try to preserve some of the custom error text, specifically

- `GET_REPORT_FAILED`
- `GET_REPORTS_BY_STATE_FAILED`
- `SET_REPORT_FAILED`

Those are the ones that are being set by various functions in the ReportProvider. Could we sync up with design/product to create versions of the new error (with link, etc) for each of those use cases? It might be as easy as keeping the first sentence for each of these, and the rest of the description can be the same as the generic error. The issue is that if we make all of the error messages the same, we eliminate our first entry point for debugging, which is to determine which function generated the error we see in the UI.

PR #489, opened 2024-03-08
Remove date-fns and date-fns-tz from app-api

benmartin-coforma:
### Description
We can skip a package import (and all of the resulting Snyk/dependabot noise) by invoking the built-in `Intl.DateTimeFormat` functions. This function has been fully supported in NodeJS since 13.0.

This upgrade should be safe because we are only changing the API - therefore we are not relying on browser versions or users' machine configurations. If the unit tests pass in Github's runner, and reports are created with the correct date format in this branch's ephemeral environment, we will know we are regression-free.

### Related ticket(s)
n/a

---
### How to test
* Log in to this branch's cloudfront environment as a state user
* Create a Work Plan
* Examine the payload in the network tab when you fetch this work plan. The top-level `reportYear` property should be 2024.
* For bonus points, fill out & submit that Work Plan, then examine the fieldData in the DB or in the network tab. The top-level `reportSubmissionDate` property should be properly formatted with MM/dd/yyyy.

### Important updates

---
### Author checklist

### DIFF
--- a/services/app-api/package.json
+++ b/services/app-api/package.json
@@ -33,8 +33,6 @@
   "dependencies": {
     "aws-jwt-verify": "^3.1.0",
     "aws-sdk": "^2.1525.0",
-    "date-fns": "^2.26.0",
-    "date-fns-tz": "^1.2.2",
     "dompurify": "^2.4.0",
     "jsdom": "^20.0.0",
     "jwt-decode": "^3.1.2",
--- a/services/app-api/tsconfig.json
+++ b/services/app-api/tsconfig.json
@@ -10,8 +10,6 @@
     "outDir": ".build",
     "rootDir": "./",
     "resolveJsonModule": true,
-    // This module resolution strategy avoids issues with the types of date-fns-tz
-    "moduleResolution": "node",
     "lib": ["ES2022"],
     "module": "node16",
     "target": "ES2022"
--- a/services/app-api/utils/time/time.test.ts
+++ b/services/app-api/utils/time/time.test.ts
@@ -1,5 +1,26 @@
 import { ReportType } from "../types";
-import { calculateDueDate, calculatePeriod, isLeapYear } from "./time";
+import {
+  calculateDueDate,
+  calculatePeriod,
+  convertDateUtcToEt,
+  isLeapYear,
+} from "./time";
+
+describe("Test convertDateUtcToEt", () => {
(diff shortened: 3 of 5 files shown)


benmartin-coforma:
I confirmed in this branch's environment, the fieldData `reportSubmissionDate` was correctly formatted.

jessabean (approved):

karla-vm (approved):

BearHanded:
Is this a change that should happen to mcr's time logic as well?

benmartin-coforma:
> Is this a change that should happen to mcr's time logic as well? alternatively, should we move this util file into the core repo and use it in both?

In my opinion, yes! Ideally we would remove `date-fns` and `date-fns-tz` from all of our apps. But there is a wrinkle.

In the case of MFP's app-api, the only `date-fns-tz` function we were using was this date format function, which goes from UTC -> ET. Since that's trivial to do with the browser's knowledge of timezones via `Intl.DateTimeFormat` (which is indeed how `date-fns-tz` does it), we can easily remove the dependency here.

In the case of MFP's ui-src, and several other services across our various apps, we use `date-fns-tz` to go the _other_ direction: from ET -> UTC. This is decidedly _not_ a trivial operation. Ultimately `date-fns-tz` does still use `Intl.DateTimeFormat` to make this happen, and we could do the same, and I believe we _should_ do the same.

I want to write that additional ET -> UTC function, and then I'll be able to make a follow-up PR to remove `date-fns-tz` from MFP's ui-src. And potentially, corresponding PRs to our other apps.

-----

As for the question of where these `Intl.DateTimeFormat` calls should live, I'm of the somewhat-weak opinion that they should live, duplicated, within each service, within each app. That's how we've handled `time.ts` up to this point, and I don't see it being worthwhile to move it to core. I could be convinced otherwise on this point, but I would want to at least have the discussion.

-----

Finally, I'll just note that the ET -> UTC conversion is never used in a way that feels particularly important to me. For example, in MFP's ui-src, the only place we perform this conversion is while saving an admin banner, so that the start & end dates for a banner can be at specific times of day on the east coast (noon and 11:59pm, respectively (which feels wrong by itself? Should it not be from midnight to midnight?)). While I don't intend to break this behavior, I also don't think it would be a showstopper if we decided to hardcode these times to, say, UTC-4, ignoring the effects of DST in the wee hours of the night.

-----

Does any of that block this PR? No, I don't think so.

PR #491, opened 2024-03-11
Fix on TR Console Error

keeysnc:
### Description
The following error was displaying due to divs wrapping `tr` tags inside some of the PDF tables. 
`<tr> cannot appear as a child of <div> `

I removed the div's and updated the `tr` tags so the styling wouldn't change. Direct child `div` tags inside `tr` tags were converted to `td` tags.

Note: I think initially adding the div as a wrapper was to be able to wrap a whole entity section with this `key={`${reportType}${idx}`}` .  I removed the div and placed the key on the `tr` which only wraps the heading of the section. I think since the key is on each section header it should be enough (if its just an identifier ). I'm open to any thoughts on this.

### Related ticket(s)
CMDCT-https://jiraent.cms.gov/browse/CMDCT-3369

---
### How to test
1. Fill out both WP and SAR reports then check the tables in the PDF against the mockup. Styling shouldn't have changed.
2. Check the console while the WP and SAR PDFs are open and this error shouldnt exist. `<tr> cannot appear as a child of <div> ` 

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/components/export/ExportedEntityDetailsOverlaySection.tsx
+++ b/services/ui-src/src/components/export/ExportedEntityDetailsOverlaySection.tsx
@@ -1,5 +1,5 @@
 // components
-import { Box, Heading, Text } from "@chakra-ui/react";
+import { Box, Heading, Text, Td } from "@chakra-ui/react";
 import { Fragment } from "react";
 import uuid from "react-uuid";
 // components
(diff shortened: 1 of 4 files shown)


britt-mo (approved):
Tested, did not see the console error

PR #511, opened 2024-03-27
Error alert refactor

britt-mo:
### Description

Update to the error messages throughout the app-- some styling changes but the biggest change is that every error is now an object of type `ErrorVerbiage` that includes a `title` and `description`. 

Screenshot 2024-03-29 at 11 14 44 AM

Screenshot 2024-03-28 at 2 37 55 PM

### Related ticket(s)

CMDCT-3282

---
### Important updates

---
### Author checklist

---


services/ui-src/src/components/reports/ReportProvider.tsx
```diff
@@ -42,12 +48,12 @@ export const ReportContext = createContext<ReportContextShape>({
   releaseReport: Function,
   // other
   isReportPage: false as boolean,
-  errorMessage: undefined as string | undefined,
+  errorMessage: undefined as unknown as ErrorVerbiage,
```
gmrabian, line 51: 😵

services/ui-src/src/types/states.ts
```diff
@@ -22,14 +23,14 @@ export interface AdminBannerState {
   bannerData: AdminBannerData | undefined;
   bannerActive: boolean;
   bannerLoading: boolean;
-  bannerErrorMessage: string;
+  bannerErrorMessage: ErrorVerbiage;
```
jessabean, line 26: We can make this a union type like `bannerData` and allow `undefined` as an option, so we can set the error state to a falsy value...

services/ui-src/src/utils/state/useStore.ts
```diff
@@ -33,7 +34,7 @@ const bannerStore = (set: Function) => ({
   bannerData: undefined,
   bannerActive: false,
   bannerLoading: false,
-  bannerErrorMessage: "",
+  bannerErrorMessage: { title: "", description: "" },
```
jessabean, line 37: I would follow the same pattern as `bannerData` here.

karla-vm (dismissed):

services/ui-src/src/components/banners/AdminBannerProvider.tsx
```diff
@@ -41,7 +41,10 @@ export const AdminBannerProvider = ({ children }: Props) => {
       const currentBanner = await getBanner(ADMIN_BANNER_ID);
       const newBannerData = currentBanner?.Item || {};
       setBannerData(newBannerData);
-      setBannerErrorMessage("");
+      setBannerErrorMessage({
```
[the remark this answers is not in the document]
jessabean: I still see a persistent banner here. Since we are deliberately calling `setBannerErrorMessage` in the try block with error verbiage, the alert is going to show up when we're checking the store. This should probably be `setBannerErrorMessage(undefined)`

jessabean (approved):
🎉 Let's finally get this merged!!

karla-vm (approved):

PR #512, opened 2024-03-28
Upgrade AWS SDK to v3

gmrabian:
### Description

Upgrade MFP to use AWS SDK v3

This PR was greatly assisted by the efforts of @benmartin-coforma in the [upgrade for CARTS](https://github.com/Enterprise-CMCS/macpro-mdct-carts/pull/139521)

Tips for reviewing:
- Start with everything other than app-api
- Once you get to app-api start with package.json, then the utils/ folder
- Almost every change to a handler is updating how it uses the utils and new types
- Almost every change to a test is updating mocks for the new sdk

Notes:
- s3 get is different https://docs.aws.amazon.com/AmazonS3/latest/userguide/example_s3_GetObject_section.html
  - it now returns a streamed body, which we have to convert to a string and then parse. this is updated in the s3-lib with virtually no impact elsewhere (except mocks in tests)

References:
- AWS SDK v3 [docs](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/)

### Related ticket(s)

CMDCT-3076

---
### How to test

Deployed url: https://d3un5oc4ck0won.cloudfront.net/

Click around and make sure everything is ok!

I already verified that the avscan lambda still works after removing the unused libs from /uploads

### Important updates

yarn everything

---
### Author checklist

---


services/app-api/utils/dynamo/dynamodb-lib.ts
```diff
@@ -1,57 +1,62 @@
...
-      params.ExclusiveStartKey = result.LastEvaluatedKey;
-      complete = result.LastEvaluatedKey === undefined;
-    }
-    return { Items: items, Count: items.length };
+  delete: async (params: DeleteCommandInput) =>
+    await client.send(new DeleteCommand(params)),
+  put: async (params: PutCommandInput) =>
+    await client.send(new PutCommand(params)),
+  query: async (params: QueryCommandInput) =>
+    await client.send(new QueryCommand(params)),
```
benmartin-coforma, line 48: Thought: We don't have any code in the API where we couldn't drop in `queryAll` in place of `query`, and I believe any future call to `query` is more likely to be a latent bug than a deliberate choice to ignore extra result pages.

I would vote to consolidate the two methods: to either remove `query` and replace the two remaining calls (both in formTemplates.ts) with `queryAll`, or to rename `queryAll` to `query` (making it the default behavior).

services/app-api/utils/formTemplates/formTemplates.test.ts
```diff
@@ -62,8 +70,19 @@ const generateReportHash = (
...
+    dynamoClientMock
+      .on(QueryCommand)
+      // mocked once for search by hash
+      .resolvesOnce({
+        Items: [],
+      })
+      // mocked again for search for latest report
+      .resolvesOnce({
+        Items: [],
+      });
```
benmartin-coforma, line 85: Idea: this mocking library allows targeting calls by argument, so we _could_ make use of that like so:
```suggestion
    dynamoClientMock
      .on(QueryCommand, {
        KeyConditionExpression: "reportType = :reportType AND md5Hash = :md5Hash",
      })
      .resolves({ Items: [] });
    dynamoClientMock
      .on(QueryCommand, {
        KeyConditionExpression: "reportType = :reportType",
        ScanIndexForward: false,
      })
      .resolves({ Items: [] });
```

This is a small win, in that the mocking is more explicit. But it also comes at a small cost, in that the mocking is more verbose. Occasionally jest's `.mockReturnValueOnce` and friends rely on inappropriate knowledge of the order in which the function under test makes its calls... but that's not really the case here, and you could argue that we would only be swapping it for inappropriate knowledge of the calls' arguments.

Stepping back to a bigger picture, it would be nice if the two query methods (`getTemplateVersionByHash` and `getNewestTemplateVersion`) were elsewhere, so that they could be mocked by name here. That would both avoid any inappropriate knowledge and prevent the need to directly mock the dynamo client for this business-logic test. But that is also a bigger refactor than we probably want to incorporate into this PR.

tl;dr: This is fine but not great. You could change it if you wanted, to something else... that would also be just fine and not great. 🤷‍♂️

services/app-api/utils/s3/s3-lib.ts
benmartin-coforma, line 1: Praise: I like the rearranged exports in here. `getFieldDataKey` and `getFormTemplateKey` still feel a bit awkward next to the S3 client functions, but I'm not sure how I'd reorganize those at this time.

services/app-api/utils/types/other.ts
benmartin-coforma, line 1: Goodbye, `DynamoGet` and friends. We never needed you and we won't miss you. o7

services/app-api/handlers/reports/release.test.ts
```diff
@@ -36,13 +37,22 @@ const releaseEvent: APIGatewayProxyEvent = {
...
   });
 
   test("Test release report passes with valid data", async () => {
     mockAuthUtil.hasPermissions.mockReturnValueOnce(true);
-    mockDocumentClient.get.promise.mockReturnValueOnce({
+    // s3 mocks
+    const s3GetSpy = jest.spyOn(s3Lib, "get");
+    s3GetSpy
+      .mockResolvedValueOnce(mockReportJson)
+      .mockResolvedValueOnce(mockReportFieldData);
```
benmartin-coforma, line 51: Thought: This is an instance of what I alluded to in an earlier comment, where I think it would be preferable to mock the different return values based on different parameters (or possibly, to move the actual queries elsewhere, to be mocked by name here).

But of course, that's much harder to do here, because jest doesn't (afaik) provide such an easy route to match mocks to parameters. So this is totally fine as-is for now.

benmartin-coforma (dismissed):
Very nice work. I have some thoughts and suggestions but they range from minor to inconsequential. I would honestly be fine to merge this as-is.

services/app-api/utils/formTemplates/formTemplates.test.ts
gmrabian: Thanks for calling this out. I think I'll leave it for now, knowing that all of the code surrounding this could stand a refactor.

services/app-api/utils/s3/s3-lib.ts
gmrabian: Yep not sure how best to re-org and I see it as outside of scope here

benmartin-coforma (approved):

PR #529, opened 2024-04-11
Prevent submission of forms when the API would reject them

benmartin-coforma:
### Description
* State users are allowed to make changes to report data, and admin users are not.
* Every report page is a form for report data, and the report page footer contains a submit button for that form
* Currently, when an admin user navigates from page to page, they are continually submitting and re-submitting those forms
   * This fails because the API recognizes that admin users are not allowed to PUT to report endpoints
   * It also fails because the API endpoint for updating reports contains a state abbreviation, pulled from the user's data. Since admin users aren't associated with a specific state, this URL ends up malformed.
* This PR replaces the footer's submit button with a simple navigation button, when it detects that the user is an admin
   * Or when it detects that the form would not be editable for other reasons, such as the report having been submitted.

### Related ticket(s)
CMDCT-3533

---
### How to test
* Log in as an admin
* Enter any report, such as a workplan
* Open the network tab, and clear it
* Click the "Continue" button at the bottom of the page
* Note that the traffic does _not_ contain a failed PUT to `api/WP/undefined/abc-123`

### Notes
n/a

---
### Pre-review checklist

---
#### Review

#### Security

### DIFF
--- a/services/ui-src/src/components/reports/ReportPageFooter.tsx
+++ b/services/ui-src/src/components/reports/ReportPageFooter.tsx
@@ -28,7 +28,7 @@ export const ReportPageFooter = ({
    */
   const reportWithSubmittedStatus = report?.status === ReportStatus.SUBMITTED;
   const { userIsAdmin, userIsEndUser } = useStore().user ?? {};
-  const formIsDisabled = 
+  const formIsDisabled =
     (userIsAdmin && !form?.editableByAdmins) ||
     (userIsEndUser && reportWithSubmittedStatus) ||
     !editable ||
--- a/services/ui-src/src/components/reports/ReportPageFooter.tsx
+++ b/services/ui-src/src/components/reports/ReportPageFooter.tsx
@@ -3,7 +3,7 @@ import { useNavigate } from "react-router-dom";
 import { Box, Button, Flex, Image, Spinner } from "@chakra-ui/react";
 // utils
 import { parseCustomHtml, useFindRoute, useStore } from "utils";
-import { AnyObject, FormJson } from "types";
+import { AnyObject, FormJson, ReportStatus } from "types";
 // assets
 import nextIcon from "assets/icons/icon_next_white.png";
 import previousIcon from "assets/icons/icon_previous_blue.png";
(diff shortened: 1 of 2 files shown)


services/ui-src/src/components/reports/ReportPageFooter.tsx
```diff
@@ -15,14 +15,25 @@ export const ReportPageFooter = ({
...
   );
   const hidePrevious = previousRoute === "/wp" || previousRoute === "/sar";
 
+  /*
+   * By default, the Continue button submits the form AND navigates.
+   * If the form is disabled, the button will navigate only.
+   */
+  const reportWithSubmittedStatus = report?.status === ReportStatus.SUBMITTED;
+  const { userIsAdmin, userIsEndUser } = useStore().user ?? {};
+  const formIsDisabled =
```
jessabean, line 31: While we are in here, is it worth trying to write a test for this? I.e., ensure that the continue button calls the useNavigate hook if the form is disabled, instead of a submit? 

Not a blocker, because sometimes tests are more complex than it's worth, but wondering how you feel about covering our behinds for this scenario. 😆

jessabean (approved):

gmrabian (dismissed):
this looks fine. i'm wondering if there is an approach where we tighten up the logic around `editable` such that that one field could determine all of this (and other locations)

services/ui-src/src/components/reports/ReportPageFooter.tsx
benmartin-coforma: Done! Thanks for the nudge :)

benmartin-coforma:
> this looks fine. i'm wondering if there is an approach where we tighten up the logic around editable such that that one field could determine all of this (and other locations)

That's a good thought, and I did investigate a bit. Right now the logic lives in about two spots: Form.tsx and now here.
* I considered making a utility function that could be called by both, but I think it would be about 50% parameters by weight, so I don't love that.
* I considered putting it into the store, as some sort of computed - I'm assuming Zustand supports that? But the store is currently very cleanly divided between report, user, banner, and entity; this would cross those lines. Moreover, the logic involves the currently-rendering `form`, which is not in the store.
   * We do only reference the form to get the `editableByAdmins` flag, which isn't strictly necessary here (it being the ReportPageFooter) - but it is necessary within Form.tsx, because it renders both report forms and banner forms.

I think between the options I'd prefer to put this logic in the store; "Can the current user edit the current form" sounds like a question worth answering from the global app state. But because that would probably also entail Storing the current form, I'd prefer not to take that on until we see another instance of code duplication.

CCing @karla-vm, in case she has anything to add.

jessabean (approved):

gmrabian (approved):

PR #538, opened 2024-04-16
Disable Funding Source Radio List 

keeysnc:
### Description
[Dev link](https://d158nmtni3yrdc.cloudfront.net/)
Disabled editable capabilities on Funding Source RadioList when a state user is in a 'copy over' report. 

### Related ticket(s)

CMDCT-[ cmdct-3487-disablefs](https://jiraent.cms.gov/browse/CMDCT-3487)

---
### How to test
1. Create a WP report 
2. Create a copy over report 
3. Go into report and select an an initiative from State or Territory-Specific Initiatives
4. Edit III. Funding sources and you should only be able to access the quarters ( radio buttons should all be disabled )

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/ui-src/src/components/fields/ChoiceListField.test.tsx
+++ b/services/ui-src/src/components/fields/ChoiceListField.test.tsx
@@ -3,14 +3,8 @@ import { axe } from "jest-axe";
 //components
 import { useFormContext } from "react-hook-form";
 import { ChoiceListField, ReportContext } from "components";
-import {
-  mockReportMethods,
-  mockReportsByState,
-  mockWPCopiedReport,
-  mockWpReportContext,
-} from "../../utils/testing/mockReport";
+import { mockWpReportContext } from "../../utils/testing/mockReport";
 import { ReportStatus } from "../../types";
-import { genericErrorContent } from "verbiage/errors";
 
 const mockTrigger = jest.fn().mockReturnValue(true);
 const mockSetValue = jest.fn();
@@ -179,18 +173,6 @@ const RadioComponent = (
     validateOnRender={true}
   />
 );
-
-const mockCopiedWpReportContext = {
-  ...mockReportMethods,
-  report: mockWPCopiedReport,
-  reportsByState: mockReportsByState,
-  copyEligibleReportsByState: mockReportsByState,
-  errorMessage: {
-    title: "We've run into a problem",
-    description: genericErrorContent,
-  },
-  lastSavedTime: "2:00 PM",
-};
 
 describe("Test ChoiceListField component rendering", () => {
   it("ChoiceList should render a normal Radiofield that doesn't have children", () => {
@@ -877,29 +859,6 @@ describe("ChoiceListField handles triggering validation", () => {
     mockGetValues(undefined);
     render(choiceListFieldValidateOnRenderComponent);
     expect(mockTrigger).toHaveBeenCalled();
-  });
-});
-
-describe("Test ChoiceList disabling", () => {
-  const disabledChoiceListFieldComponent = (
-    <ReportContext.Provider value={mockCopiedWpReportContext}>
-      <ChoiceListField
-        choices={mockChoices}
-        label="Checkbox example"
-        name="checkboxField"
-        type="checkbox"
-        disabled={true}
-        validateOnRender
-      />
-    </ReportContext.Provider>
-  );
-
-  test("Should disable radiofield when state user is in copy over report", async () => {
-    mockGetValues(undefined);
-    render(disabledChoiceListFieldComponent);
-    const radioBtn = document.querySelector(".ds-c-choice-wrapper")
-      ?.children[0];
-    expect(radioBtn).toBeDisabled();
   });
 });
 
--- a/services/ui-src/src/components/fields/ChoiceListField.test.tsx
+++ b/services/ui-src/src/components/fields/ChoiceListField.test.tsx
@@ -3,8 +3,14 @@ import { axe } from "jest-axe";
 //components
 import { useFormContext } from "react-hook-form";
 import { ChoiceListField, ReportContext } from "components";
-import { mockWpReportContext } from "../../utils/testing/mockReport";
+import {
+  mockReportMethods,
+  mockReportsByState,
+  mockWPCopiedReport,
+  mockWpReportContext,
+} from "../../utils/testing/mockReport";
 import { ReportStatus } from "../../types";
+import { genericErrorContent } from "verbiage/errors";
 
 const mockTrigger = jest.fn().mockReturnValue(true);
 const mockSetValue = jest.fn();
@@ -173,6 +179,18 @@ const RadioComponent = (
     validateOnRender={true}
   />
 );
+
+const mockCopiedWpReportContext = {
+  ...mockReportMethods,
+  report: mockWPCopiedReport,
+  reportsByState: mockReportsByState,
+  copyEligibleReportsByState: mockReportsByState,
+  errorMessage: {
+    title: "We've run into a problem",
+    description: genericErrorContent,
+  },
+  lastSavedTime: "2:00 PM",
+};
 
 describe("Test ChoiceListField component rendering", () => {
   it("ChoiceList should render a normal Radiofield that doesn't have children", () => {
@@ -859,6 +877,29 @@ describe("ChoiceListField handles triggering validation", () => {
     mockGetValues(undefined);
     render(choiceListFieldValidateOnRenderComponent);
     expect(mockTrigger).toHaveBeenCalled();
+  });
+});
+
+describe("Test ChoiceList disabling", () => {
+  const disabledChoiceListFieldComponent = (
+    <ReportContext.Provider value={mockCopiedWpReportContext}>
+      <ChoiceListField
+        choices={mockChoices}
+        label="Checkbox example"
+        name="checkboxField"
+        type="checkbox"
+        disabled={true}
+        validateOnRender
+      />
+    </ReportContext.Provider>
+  );
+
+  test("Should disable radiofield when state user is in copy over report", async () => {
+    mockGetValues(undefined);
+    render(disabledChoiceListFieldComponent);
+    const radioBtn = document.querySelector(".ds-c-choice-wrapper")
+      ?.children[0];
+    expect(radioBtn).toBeDisabled();
   });
 });
 
--- a/services/ui-src/src/components/fields/RadioField.tsx
+++ b/services/ui-src/src/components/fields/RadioField.tsx
@@ -2,6 +2,9 @@
 import { Box } from "@chakra-ui/react";
 import { ChoiceListField } from "components";
 // utils
+import { useStore } from "utils";
+
+// types
 import { ChoiceFieldProps } from "types";
 
 export const RadioField = ({
@@ -11,6 +14,13 @@ export const RadioField = ({
   sxOverride,
   ...props
 }: ChoiceFieldProps) => {
+  const { userIsEndUser } = useStore().user ?? {};
+  const { report } = useStore();
+
+  const isEditableRadioFields = () => {
+    return label === "Funding source:" && userIsEndUser && report?.isCopied;
+  };
+
   return (
     <Box sx={{ ...sx, ...sxOverride }}>
       <ChoiceListField
@@ -19,6 +29,7 @@ export const RadioField = ({
         label={label}
         choices={choices}
         {...props}
+        disabled={isEditableRadioFields()}
       />
     </Box>
   );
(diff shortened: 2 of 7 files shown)


benmartin-coforma (changes requested):
I don't think this is the right place to make this change. I think I would do the logic earlier, during `updateRenderFields`. We already have a fair bit of what you might call form pre-processing logic in there... manipulating target populations and so on. So I think this would fit in fairly well.

I would probably create a new function within forms.ts, called something like `disableCopiedFundingSources(report: ReportShape, fields: (FormField | FormLayoutElement)[])`. It would find the field(s?) we're trying to target, probably by ID, and set `field.props.disabled` to `true`. That prop is the same way we specify that auto-populated fields are disabled, so there's precedent there.

I don't _think_ you'll need to touch RadioField.tsx at all. It already passes `...props` through to the ChoiceListField, and presumably that gets passed through to the actual input elements.

It would be more awkward to access the user store from within forms.ts, but happily we won't need to! Only state users are allowed to make changes to reports to begin with, so I don't think any role-checking will be needed.

keeysnc:
> I don't think this is the right place to make this change. I think I would do the logic earlier, during `updateRenderFields`. We already have a fair bit of what you might call form pre-processing logic in there... manipulating target populations and so on. So I think this would fit in fairly well.
> 
> I would probably create a new function within forms.ts, called something like `disableCopiedFundingSources(report: ReportShape, fields: (FormField | FormLayoutElement)[])`. It would find the field(s?) we're trying to target, probably by ID, and set `field.props.disabled` to `true`. That prop is the same way we specify that auto-populated fields are disabled, so there's precedent there.
> 
> I don't _think_ you'll need to touch RadioField.tsx at all. It already passes `...props` through to the ChoiceListField, and presumably that gets passed through to the actual input elements.
> 
> It would be more awkward to access the user store from within forms.ts, but happily we won't need to! Only state users are allowed to make changes to reports to begin with, so I don't think any role-checking will be needed.

Thanks for reviewing! Yea that seems like a good idea. I took a look at forms.ts and it looks so cluttered lol Makes sense since the forms have a lot of logic though. Definitely had a slight feeling RadioField should be untouched but, then I did it.

benmartin-coforma:
Awesome! Thanks for making that change, I'm glad it worked.

I do still have some nits to pick though, and I apologize for the wall of text here...
* Within disableCopiedFundingSources, I'd recommend a function other than `.map`, for semantic reasons. Seeing the word "map" tells me (and our app's future maintainers) that we're going to take a sequence and transform it, element-by-element. Here, it's only being used to loop, so it could be directly swapped for `.forEach` to improve clarity.
   * In general, the callback function passed to `.map` should always return something. Otherwise, it's not really mapping anything.
* But, it's my opinion that `.forEach` is a worse choice than `for (... of ...) { }`. It's convenient that you can hang `.forEach` on the end of an array to loop over it, but it doesn't support early returns, `break`, `continue`, or `await`. So I always prefer to use `for..of`
* Within that loop, the value of `report?.isCopied` doesn't change, so it could be lifted out: If the report isn't copied, we can early return from this function and skip the loop.
* But, this function really doesn't need a loop at all. We know that every field in a form will have unique IDs, so we know that the number of fields with a specific ID will be either 0 or 1. This is the use case for `.find`; the code might look like `const fundingSourceField = fields.find((f) => f.id === "fs_wpt");`, followed by `if (fundingSourceField) { fundingSourceField.props = {...myField.props, disabled: true} }`
* But we're searching for a specific field by ID, so we know exactly what it's going to look like when we find it. Specifically, we know that it's going to have props - see line 1139 in wp.json to confirm that. So we can skip the spread operator in favor of a direct assignment: `fundingSourceField.props!.disabled = true;`
* Since we know that the field we're targeting is the first in its form, we _could_ go so far as to skip the `.find`, and just check the ID on `fields[0]`. I don't think I would go that far though; we know that our field IDs are permanent but we don't usually make that assumption about the _order_ of the fields. Maybe someday there will be another question before it, or even just a form layout element.
* Personally I don't think the `structuredClone` is necessary here. We're performing an idempotent operation, at the last minute before the form is rendered, so I would be fine just altering the field in place.
* I do think the parameter order is important. The new function is being called from `updateRenderFields(report, fields)`, so it should also take `report` as its first parameter and `fields` as its second.
* Jumping back out to `updateRenderFields`... This one is even more opinion-based, but I would prefer an old-fashioned `if` over the ternary operator here. I'm having trouble describing why... Our new function is a significant enough operation that it feels like it "deserves" the whole `if (condition) { doThing(); }` ceremony. If that makes sense.
* But I don't think we want the conditional here at all. For one, because the `updateRenderFields` function is already complicated enough; when a function grows to this size I think it's better to have it delegate that kind of logic out to the smaller functions it calls. For two, because this condition assumes that this field will always be the first in the form, which I pointed out earlier seems like a bad idea.
* Finally, the comment: "disable funding source radio buttons on copy over reports". This is very close to being a repeat of our new function name - So maybe we need to give the function a more descriptive name? It could very well be called something like`disableFundingSourceRadioButtonsForCopiedReports(report, fields)`, and we could drop the comment.
* Alternatively, if we wanted to give a comment with more of the "why" and less of the "what", that explanation would be perfectly at home in a doc comment on top of the new function. And in fact, that's a good idea here, because "why are we even doing this" was my very first question.
   
I'm not going to make you go through that wall of text and make each edit one by one; here's the sum of all my suggestions:

```ts
/**
 * Prevent users from changing a funding source's type in a copied report.
 * 
 * If the user copied work plan A to make work plan B, we assume that the
 * data in A was correct. When editing B, they can adjust amounts and
 * projections, but they can't make fundamental changes to the source.
 */
export const disableCopiedFundingSources = (
  report: ReportShape,
  fields: (FormField | FormLayoutElement)[]
) => {
  if (!report?.isCopied) {
    return;
  }

  const fundingSourceField = fields.find(
    (field) => field.id === "fundingSources_wpTopic"
  );

  if (fundingSourceField) {
    fundingSourceField.props!.disabled = true;
  }
};

export const updateRenderFields = (
  report: ReportShape,
  fields: (FormField | FormLayoutElement)[]
) => {
  disableCopiedFundingSources(report, fields);
  ...
```

This would go along with these changes to the tests:
```ts
    describe("disableCopiedFundingSources", () => {
      it("should disable choicelist on copy over report - funding source section", () => {
        const fields: FormField[] = [structuredClone(mockFundingSourceFormField)];
        disableCopiedFundingSources(
          mockWPCopiedReport,
          fields
        );
        expect(fields[0].props?.disabled).toBe(true);
      });

      it("should not disable funding sources for non-copied reports", () => {
        const fields: FormField[] = [structuredClone(mockFundingSourceFormField)];
        disableCopiedFundingSources({} as ReportShape, fields);
        expect(fields[0].props?.disabled).toBeFalsy();
      });

      it("should not disable fields other than funding sources", () => {
        const fields: FormField[] = [structuredClone(mockFormField)];
        disableCopiedFundingSources(mockWPCopiedReport, fields);
        expect(fields[0].props?.disabled).toBeFalsy();
      });
    });
```

keeysnc:
> Awesome! Thanks for making that change, I'm glad it worked.
> 
> I do still have some nits to pick though, and I apologize for the wall of text here...
> 
> * Within disableCopiedFundingSources, I'd recommend a function other than `.map`, for semantic reasons. Seeing the word "map" tells me (and our app's future maintainers) that we're going to take a sequence and transform it, element-by-element. Here, it's only being used to loop, so it could be directly swapped for `.forEach` to improve clarity.
>   
>   * In general, the callback function passed to `.map` should always return something. Otherwise, it's not really mapping anything.
> * But, it's my opinion that `.forEach` is a worse choice than `for (... of ...) { }`. It's convenient that you can hang `.forEach` on the end of an array to loop over it, but it doesn't support early returns, `break`, `continue`, or `await`. So I always prefer to use `for..of`
> * Within that loop, the value of `report?.isCopied` doesn't change, so it could be lifted out: If the report isn't copied, we can early return from this function and skip the loop.
> * But, this function really doesn't need a loop at all. We know that every field in a form will have unique IDs, so we know that the number of fields with a specific ID will be either 0 or 1. This is the use case for `.find`; the code might look like `const fundingSourceField = fields.find((f) => f.id === "fs_wpt");`, followed by `if (fundingSourceField) { fundingSourceField.props = {...fundingSourceField.props, disabled: true} }`
> * But we're searching for a specific field by ID, so we know exactly what it's going to look like when we find it. Specifically, we know that it's going to have props - see line 1139 in wp.json to confirm that. So we can skip the spread operator in favor of a direct assignment: `fundingSourceField.props!.disabled = true;`
> * Since we know that the field we're targeting is the first in its form, we _could_ go so far as to skip the `.find`, and just check the ID on `fields[0]`. I don't think I would go that far though; we know that our field IDs are permanent but we don't usually make that assumption about the _order_ of the fields. Maybe someday there will be another question before it, or even just a form layout element.
> * Personally I don't think the `structuredClone` is necessary here. We're performing an idempotent operation, at the last minute before the form is rendered, so I would be fine just altering the field in place.
> * I do think the parameter order is important. The new function is being called from `updateRenderFields(report, fields)`, so it should also take `report` as its first parameter and `fields` as its second.
> * Jumping back out to `updateRenderFields`... This one is even more opinion-based, but I would prefer an old-fashioned `if` over the ternary operator here. I'm having trouble describing why... Our new function is a significant enough operation that it feels like it "deserves" the whole `if (condition) { doThing(); }` ceremony. If that makes sense.
> * But I don't think we want the conditional here at all. For one, because the `updateRenderFields` function is already complicated enough; when a function grows to this size I think it's better to have it delegate that kind of logic out to the smaller functions it calls. For two, because this condition assumes that this field will always be the first in the form, which I pointed out earlier seems like a bad idea.
> * Finally, the comment: "disable funding source radio buttons on copy over reports". This is very close to being a repeat of our new function name - So maybe we need to give the function a more descriptive name? It could very well be called something like`disableFundingSourceRadioButtonsForCopiedReports(report, fields)`, and we could drop the comment.
> * Alternatively, if we wanted to give a comment with more of the "why" and less of the "what", that explanation would be perfectly at home in a doc comment on top of the new function. And in fact, that's a good idea here, because "why are we even doing this" was my very first question.
> 
> I'm not going to make you go through that wall of text and make each edit one by one; here's the sum of all my suggestions:
> 
> ```ts
> /**
>  * Prevent users from changing a funding source's type in a copied report.
>  * 
>  * If the user copied work plan A to make work plan B, we assume that the
>  * data in A was correct. When editing B, they can adjust amounts and
>  * projections, but they can't make fundamental changes to the source.
>  */
> export const disableCopiedFundingSources = (
>   report: ReportShape,
>   fields: (FormField | FormLayoutElement)[]
> ) => {
>   if (!report?.isCopied) {
>     return;
>   }
> 
>   const fundingSourceField = fields.find(
>     (field) => field.id === "fundingSources_wpTopic"
>   );
> 
>   if (fundingSourceField) {
>     fundingSourceField.props!.disabled = true;
>   }
> };
> 
> export const updateRenderFields = (
>   report: ReportShape,
>   fields: (FormField | FormLayoutElement)[]
> ) => {
>   disableCopiedFundingSources(report, fields);
>   ...
> ```
> 
> This would go along with these changes to the tests:
> 
> ```ts
>     describe("disableCopiedFundingSources", () => {
>       it("should disable choicelist on copy over report - funding source section", () => {
>         const fields: FormField[] = [structuredClone(mockFundingSourceFormField)];
>         disableCopiedFundingSources(
>           mockWPCopiedReport,
>           fields
>         );
>         expect(fields[0].props?.disabled).toBe(true);
>       });
> 
>       it("should not disable funding sources for non-copied reports", () => {
>         const fields: FormField[] = [structuredClone(mockFundingSourceFormField)];
>         disableCopiedFundingSources({} as ReportShape, fields);
>         expect(fields[0].props?.disabled).toBeFalsy();
>       });
> 
>       it("should not disable fields other than funding sources", () => {
>         const fields: FormField[] = [structuredClone(mockFormField)];
>         disableCopiedFundingSources(mockWPCopiedReport, fields);
>         expect(fields[0].props?.disabled).toBeFalsy();
>       });
>     });
> ```

@benmartin-coforma This is not a nit lol But it all sounds valid. Read the wall =)

benmartin-coforma (dismissed):
I really appreciate your patience as I gave you the runaround on this 😅

keeysnc:
> I really appreciate your patience as I gave you the runaround on this 😅

All good, lol It was good insight.

jessabean (approved):
Works great locally. ~FYI...it did _not_ seem to work as expected on the cloudfront URL. We may need to investigate that.~

Has this gotten sign off from design?

It has not, I'll send it off to them.

benmartin-coforma (approved):
As co-author, I approve 👍

gmrabian (approved):

PR #540, opened 2024-04-17
One default target population must be applicable

gmrabian:
Deployed url: https://d3krpqx3uqolmw.cloudfront.net/

### Description

Expected behavior: If all default target population drawers are complete and they are all marked 'No' (not applicable) then there is an error message indicating at least one must be applicable, and the page is considered incomplete on Review and Submit.

Added conditional logic to cypress tests to ensure we are marking target populations as applicable. Shoutout to Nick for writing an incredible cypress tests framework and shoutout to Ajaita for helping me use it properly!

I also removed all our `data-cy` logic in favor of better matchers 😁 

### Related ticket(s)

CMDCT-3488

---
### How to test

- Create a WP
- Go to Transition Benchmarks and select 'No' for all of the four default populations
- Verify that an error pops up on the page and the Review and Submit indicates that page as incomplete
- Change any number of populations to 'Yes' and verify the error goes away and the page is complete
- Verify that adding more pops, toggling defaults back and forth, etc. all work as expected

---
### Pre-review checklist

---
#### Review

---
### Feedback requested
Please advise if you think this could be better! These were the best locations I could find to make the updates.
Testing this in the ModalDrawerReportPage is really complex to set up. Open to ideas.


karla-vm (approved):
Works like a charm ✨ great idea cleaning up the target populations into constants and thanks for removing those `data-cy` identifiers!

One thing (non-blocking, just curious): do we want to ask Design to update the alert verbiage to explicitly say that it's the default target populations that are required? I added an applicable `Other` target population and the alert still shows up (which is correct, but verbiage is a bit confusing)

Screenshot 2024-04-23 at 10 53 37 AM

services/ui-src/src/components/reports/ModalDrawerReportPage.tsx
```diff
@@ -178,11 +189,46 @@ export const ModalDrawerReportPage = ({ route, validateOnRender }: Props) => {
     headRow: ["", "", ""],
   };
 
+  const defaultsPopsNotSelected = (targetPopulations: AnyObject[]) => {
+    const defaultPopulationNames = getDefaultTargetPopulationNames();
```
jessabean, line 193: Probably gonna have to write some tests for this component's changes to make code climate happy 😭  Let me know if you want help with that...

jessabean (approved):
✅ Did a regression check on the Define Initiative section as well, making sure target populations show as expected even if not applicable

Changes look good to me! I think we will need some test coverage though. 🌧️

gmrabian:
> Works like a charm ✨ great idea cleaning up the target populations into constants and thanks for removing those `data-cy` identifiers!
> 
> One thing (non-blocking, just curious): do we want to ask Design to update the alert verbiage to explicitly say that it's the default target populations that are required? I added an applicable `Other` target population and the alert still shows up (which is correct, but verbiage is a bit confusing)
> 
> Screenshot 2024-04-23 at 10 53 37 AM

Good idea! They did already approve but let me run that idea by them

karla-vm (approved):

jessabean (approved):

PR #543, opened 2024-04-17
clean up non-number report period

gmrabian:
Deployed env: https://d1evf704f7yu4c.cloudfront.net/

### Description

- Remove `reportPeriod` from create payload
- Remove `reportPeriod` from field data schema
- Tighten up reportPeriod equalities to type check

**Feedback** do we need to modify this [add/edit SAR definition](https://github.com/Enterprise-CMCS/macpro-mdct-mfp/blob/main/services/ui-src/src/forms/addEditSarReport/addEditSarReport.json#L32-L34)? I think our edit buttons were disabled anyway 🤔 

### Related ticket(s)

CMDCT-3477

---
### How to test

Test WP and SAR creation.
No change to app functionality. No new errors.
I created some reports on `stateuser1` if you want to verify any data from those

---
### Pre-review checklist

### DIFF
--- a/services/app-api/handlers/reports/create.ts
+++ b/services/app-api/handlers/reports/create.ts
@@ -118,8 +118,7 @@ export const createReport = handler(
     const { metadata: unvalidatedMetadata, fieldData: unvalidatedFieldData } =
       unvalidatedPayload;
 
-    const creationValidationJson = {
-      reportPeriod: "text",
+    const creationFieldDataValidationJson = {
       stateName: "text",
       stateOrTerritory: "text",
       submissionCount: "number",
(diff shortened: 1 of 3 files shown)


keeysnc (approved):
Tested in cloudfront and everything looks good!

jessabean (approved):

jessabean:
re: SAR definition...isn't that the text input that appears here? I think we need to keep it, unless I'm misunderstanding the question.

Screenshot 2024-04-18 at 4 01 22 PM

gmrabian:
> re: SAR definition...isn't that the text input that appears here? I think we need to keep it, unless I'm misunderstanding the question.
> Screenshot 2024-04-18 at 4 01 22 PM

thanks for showing me! yes i agree it should stay.

PR #560, opened 2024-04-22
Validate path parameters with strong typing

benmartin-coforma:
### Description
At the top of every API handler function, validate that
1. All the appropriate path parameters were included
2. Those parameters have valid values

This allows us to refer to the parameters throughout the handler without further validation or type casting.

### Related ticket(s)
N/A

---
### How to test
https://d2uyxh53bzozjt.cloudfront.net/
* The only behavioral change is slightly improved logging in the case of invalid parameters. To observe this:
   1. Log in to the app
   2. Open the network tab (and wipe it to reduce noise)
   3. Perform _some request_ - any query or modification to a report
   4. Right click that request in the network tab, and `Copy > Copy as fetch`
   5. Switch over to the browser console
   6. Paste the request (it will be quite large because it includes all headers)
   7. BEFORE you press enter, modify the path. Alter the report type or state to some invalid value.
   8. Send the request from your browser console
   9. Observe the warning in the API logs concerning the changed parameter
   10. You should receive a 400 response - although in almost all cases you already would have.
      * (I believe some handlers would previously have attempted to use the invalid value, eventually causing a 404 or 500)
* Note that the "Invalid report ID in path" error is impossible to trigger in normal circumstances:
   * The new validation function only requires that it be a string
   * If it's not a string, the only thing it could be is undefined
   * If it's not provided, serverless will not (or at least, should not) match the request to the same endpoint
   * So the relevant validation will not run
* Strictly speaking, since all endpoints were touched, they should all be regression tested.
   * Cypress will have covered most of this for us; Create, Fetch, and Update are gimmes.
   * If we have a successful SAR test, then Submit and Approve will have been tested as well
   * That leaves Archive and Release.
   * Note that the parameter _validation_ need not be tested on all endpoints; it's the same function no matter where you call it from. We just need to ensure that none of the handlers are straight-up broken.

### Notes
n/a

---
### Pre-review checklist
- ~[ ] I have performed a self-review of my code~

---
#### Review

#### Security


gmrabian (approved):
Looks good!
I tested all the lambdas in the deployed environment.
I modified the network requests and didn't see the error logs in the lambda, but maybe I missed them.

ntsummers1 (approved):
I like.

PR #564, opened 2024-04-23
Fix Users error on SAR dashboard 

keeysnc:
### Description
This should fix the SAR error banner from temporarily showing up on the SAR dashboard when a user's WP has been approved.

Screenshot 2024-04-23 at 12 44 37 PM

### Related ticket(s)

CMDCT-[https://jiraent.cms.gov/browse/CMDCT-3423]

---
### How to test
1. Fill out WP as state user
2. Approved WP as admin
3. Once report is approved go back in as a state user and view the SAR dashboard
4. The SAR error banner should not be enabled.

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/ui-src/src/components/pages/Dashboard/DashboardPage.tsx
+++ b/services/ui-src/src/components/pages/Dashboard/DashboardPage.tsx
@@ -288,12 +288,10 @@ export const DashboardPage = ({ reportType }: Props) => {
         <Image src={arrowLeftIcon} alt="Arrow left" className="returnIcon" />
         Return home
       </Link>
-      <Box sx={sx.errorMessage}>
-        {errorMessage && <ErrorAlert error={errorMessage} />}
-      </Box>
+      {errorMessage && <ErrorAlert error={errorMessage} />}
       {/* Only show SAR alert banner if the corresponding Work Plan is not approved */}
       <Box sx={sx.leadTextBox}>
-        {reportType === ReportType.SAR && !workPlanToCopyFrom === undefined && (
+        {reportType === ReportType.SAR && workPlanToCopyFrom === null && (
           <Alert
             title={sarVerbiage.alertBanner.title}
             showIcon={true}
@@ -501,9 +499,6 @@ const sx = {
     p: {
       fontSize: "16px",
     },
-  },
-  errorMessage: {
-    paddingTop: "1rem",
   },
 };
 
--- a/services/ui-src/src/components/pages/Dashboard/DashboardPage.tsx
+++ b/services/ui-src/src/components/pages/Dashboard/DashboardPage.tsx
@@ -107,6 +107,9 @@ export const DashboardPage = ({ reportType }: Props) => {
   // if a user is an admin type, use the selected state, otherwise use their assigned state
   const activeState = userIsAdmin ? adminSelectedState : userState;
 
+  const showSarAlert =
+    reportType === ReportType.SAR && !workPlanToCopyFrom && reportsToDisplay;
+
   useEffect(() => {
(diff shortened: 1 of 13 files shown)


services/ui-src/src/components/pages/Dashboard/DashboardPage.tsx
```diff
@@ -92,6 +92,9 @@ export const DashboardPage = ({ reportType }: Props) => {
   const [selectedReport, setSelectedReport] = useState<AnyObject | undefined>(
     undefined
   );
+  const [wpToCopyFrom, setWorkPlanToCopyFrom] = useState<boolean>(
```
jessabean, line 95: So, there is already a function to set the WP to copy from in the Report Provider. Which makes this state a little redundant.

I was thinking, maybe we could instead move where we currently call `setWorkPlanToCopyFrom` [in the fetch call](https://github.com/Enterprise-CMCS/macpro-mdct-mfp/blob/main/services/ui-src/src/components/reports/ReportProvider.tsx#L116). Like maybe move that line to an `else` block on [line 138](https://github.com/Enterprise-CMCS/macpro-mdct-mfp/blob/main/services/ui-src/src/components/reports/ReportProvider.tsx#L138)?

keeysnc: Good catch, I had to look through it before commenting on this but I think the undefined might be causing the issue.  `fetchReportForSarCreation(activeState)` is being called in DashboardPage (line 119) and it's not resolving right away which I think is returning an `undefined` at first which is causing the banner to temporarily flicker. 

I added a null as the default value so even if fetchReportForSarCreation doesn't resolve right away the banner wont render until a null or `workPlanToCopyFrom` object comes through. 

Open to suggestions of this solution though.

gmrabian (approved):
After approving as an admin I checked the SAR dashboard and the banner did appear for a second before disappearing.
When I logged in as a state user it didn't show up at all.

Looks good! I don't think the flicker I saw on admin is a huge concern, just calling out that the state isn't cleared right away

services/ui-src/src/components/pages/Dashboard/DashboardPage.tsx
```diff
@@ -293,16 +293,18 @@ export const DashboardPage = ({ reportType }: Props) => {
...
-          <Alert
-            title={sarVerbiage.alertBanner.title}
-            showIcon={true}
-            icon={alertIcon}
-            status={AlertTypes.ERROR}
-            description={sarVerbiage.alertBanner.body}
-            sx={sx.alertBanner}
-          />
-        )}
+        {reportType === ReportType.SAR &&
```
jessabean, line 296: This is just a nit, but when the conditional gets this long, I usually like to pull it into a variable, just to keep the template a bit cleaner.

jessabean (dismissed):
Smol nit, but this is working as expected!

services/ui-src/src/components/pages/Dashboard/DashboardPage.tsx
keeysnc: Sounds good, just updated this with a variable for the logic.

jessabean (approved):

gmrabian (commented):
When I create a SAR the error pops up. I understand that I am now locked from creating another, but from a UX perspective it looks like something went wrong by creating it

Screenshot 2024-04-29 at 11 18 59 AM

Update: Spoke with Megan and she verified this is the correct flow. Possible circle back to better the UX for this.

gmrabian (approved):

PR #569, opened 2024-04-26
3541 - Change validation for quarterly projections in WP

jessabean:
### Description
Changes validation for _quarterly projection_ "number" fields in the WP so that users can enter whole numbers OR percentages.

### Related ticket(s)

CMDCT-3541

---
### How to test
**Test link: **https://d3ozpalbymoth5.cloudfront.net/
User with plans: `stateuser@test.com` / PR / Puerto Rico

#### Work plan:
1. Create a work plan and go the the State or Territory-Specific Initiatives section.
2. Add a new initiative and in the Evaluation Plan section, click "Add objective."
3. In the Add Objective modal, select Yes for "Does the performance measure include quantitative targets?"
    - The quarterly objectives fields should allow user to enter any text, not just numbers. (e.g., "50%" should be a valid input)
    - The validation for required should still work. 
    - There should be no masking of input.
    - The styling of the input should look the same as other number fields (slightly less wide than normal text inputs).

Screenshot 2024-04-26 at 11 24 57 AM

#### SAR:

1. Create a SAR associated with the previous work plan
2. In the State or Territory-Specific Initiatives section, open a specific initiative dashboard, and select Edit for Objective Progress (be sure to pick one where quarterly projections were entered in the WP).
3. Click "Report objective progress" to open the modal.
    - The "Actual value" input fields should follow the same validation rules (styled like number inputs, but validated like text)
    - The "Target value" fields should have the values formatted the same way as they were when entered in the WP.
    
#### Regression test:
Other numerical inputs should behave the same as expected:

1. Funding source input fields are still masked as currency and only allow number inputs
2. Transition Benchmark projections should only allow whole integers or "N/A"

Screenshot 2024-04-26 at 11 29 05 AM

Screenshot 2024-04-26 at 11 19 44 AM

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/ui-src/src/utils/reports/entities.ts
+++ b/services/ui-src/src/utils/reports/entities.ts
@@ -1,5 +1,4 @@
 import { EntityShape, OverlayModalStepTypes, AnyObject } from "types";
-import { convertToThousandsSeparatedString } from "utils";
 
 const getRadioValue = (entity: EntityShape | undefined, label: string) => {
   const radioLabelValue = entity?.[label]?.[0].value;
@@ -18,10 +17,12 @@ const getRepeatedField = (
     for (const [key, value] of Object.entries(entity)) {
       if (key.includes(repeatedKey) && value) {
         const id = key.replace(repeatedKey, "").split("Q");
-        const displayValue =
-          repeatedKey === "fundingSources_quarters"
-            ? `$${value}`
-            : convertToThousandsSeparatedString(value).maskedValue;
+        let displayValue = `${value}`;
+
+        if (repeatedKey === "fundingSources_quarters") {
+          displayValue = `$${value}`;
+        }
+
         quarters.push({ id: `${id[0]} Q${id[1]}`, value: displayValue });
       }
     }
--- a/services/app-api/forms/wp.json
+++ b/services/app-api/forms/wp.json
@@ -1062,12 +1062,15 @@
                           "children": [
                             {
                               "id": "quarterlyProjections",
-                              "type": "number",
+                              "type": "text",
                               "validation": {
-                                "type": "number",
+                                "type": "text",
                                 "parentFieldName": "evaluationPlan_includesTargets",
                                 "parentOptionId": "evaluationPlan_includesTargets-7FP4jcg4jK7Ssqp3cCW5vQ",
                                 "nested": true
+                              },
+                              "props": {
+                                "className": "number-field"
                               },
                               "transformation": {
                                 "rule": "nextTwelveQuarters"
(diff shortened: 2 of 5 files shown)


benmartin-coforma (commented):
This doesn't feel great. I thought we would be using a new validation type, that required either integers or percentages. Removing the validation entirely by requiring only `text` opens the door for invalid values like
* `$50000`
* `5,0000`
* `50000.`
* `50000 (waived)`
* `same as Q2 2024`
and so on.

We had a terrific time finding and cleaning up values exactly like those from year-old MCPAR reports after discovering our validation had been too permissive over there.

services/ui-src/src/utils/reports/entities.ts
```diff
@@ -18,10 +17,12 @@ const getRepeatedField = (
...
         const id = key.replace(repeatedKey, "").split("Q");
-        const displayValue =
-          repeatedKey === "fundingSources_quarters"
-            ? `$${value}`
-            : convertToThousandsSeparatedString(value).maskedValue;
+        let displayValue = `${value}`;
+
+        if (repeatedKey === "fundingSources_quarters") {
+          displayValue = `$${value}`;
+        }
```
benmartin-coforma, line 24: This version removes thousands separators from `quarterlyProjections`, `objectiveTargets_projections_`, and `objectiveTargets_actual_`. Is it intended to affect all three, or do we still need a special case for `quarterlyProjections`?

jessabean: I believe the intention is to remove it from all of them. I believe the `objectiveTargets` fields are the SAR equivalents of `quarterlyProjections`? But either way, if it has to do with the objectives, product has been clear that we are just going to let the user enter _whatever_.

[Slack convo here](https://coforma.slack.com/archives/C06N0V30TM4/p1714147864925089), for reference

ntsummers1 (changes requested):
Have the same worry as Ben. Would definitely prefer to see a new validation type with its own tests that check for compliance for this rather than moving this to having a text validation that would allow a lot more than just whole numbers or percentages.

jessabean:
> Have the same worry as Ben. Would definitely prefer to see a new validation type with its own tests that check for compliance for this rather than moving this to having a text validation that would allow a lot more than just whole numbers or percentages.

This was a specific request from product to move to a text validation for this field, in order to avoid the extra effort to think through specific validations at this time and still meet the request from our stakeholders. Is there a concern with the technical aspects of having this field be a simple text field? Are there some dependencies with the API that might cause problems?

benmartin-coforma (approved):

ntsummers1 (approved):

ntsummers1:
(Written by Ben) For posterity: We three just had a conversation about this ticket, and Nick and I approved the PR as-is (effectively removing validation from these fields). Ideally we will be able to follow up after launch to suss out more precise business requirements and implement them, and hopefully we will be able to do so before too much crazy data finds its way into the DB :slightly_smiling_face:

PR #575, opened 2024-05-01
Add handling for internal user

braxex:
### Description

This fixes two issues:
1. the frontend of MFP does not currently account for internal users at all
2. read-only users were not being handled correctly in the admin dashboard state selection form

Changes:
- Added internal user to valid users for the frontend
- Extended read-only user check to include internal user alongside help desk user
- Changed logic related to state selection so the correct state will be selected now

### Related ticket(s)

n/a

---
### How to test

1. Log in as internal user, navigate to dashboard, select state and WP, confirm it loads correctly.
2. Repeat the same as help desk user.
3. Repeat the same as state user to confirm no lost functionality.

### Notes

n/a

---
### Pre-review checklist

---
#### Review

#### Security

---


BearHanded (approved):
Tested in the deployed branch as both internal and help desk. Created a WP as a stateuser, and was able to view it as read only for both user types.

jessabean (approved):
Thank you! 🙏🏻

PR #585, opened 2024-05-07
Add precommit check for branch names

benmartin-coforma:
### Description
This check will run every commit, ensuring that the branch name won't cause any issues* in during AWS deployment.

*Some issues are still possible. The length check may not be exactly correct; 30 characters is a guess we may need to revise later. It may also be possible to create a branch name containing a reserved word, which some AWS services can object to.

My least favorite aspect of this change is the lack of in-band feedback. If a developer creates a branch with a name that contains capital letters, for example, their first commit to that branch will get a console error like this one. It's not super informative, and there aren't any good ways to give hints as to _why_ the `Don't commit to branch` check failed. No matter how informative the comments are in .pre-commit-config.yaml, they only help if you know where to look for them.
image

But, at least that mysterious error happens locally & immediately, instead of within a GH action after a few minutes?

If we decide that this is a good idea, we should make the same change to the other apps' pre-commit configs.

### Related ticket(s)
n/a

---
### How to test
1. Check out this branch
2. Run `pre-commit install`
3. Create a new branch on top of this one, with a bad name
4. Attempt to make a commit on that branch; observe that it fails
5. Create a new branch, with a good name
6. Make a commit; observe that it succeeds

### Notes
All developers will need to run `pre-commit install` locally, after this change makes it onto their machines.

---
### Pre-review checklist

- ~[ ] I have added [thorough](https://shorturl.at/aejkF) tests, if necessary~
- ~[ ] I have performed a self-review of my code~
- ~[ ] I have manually tested this PR in the deployed cloud environment~

---
#### Review

#### Security

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,6 +3,17 @@ repos:
     rev: v4.0.1
     hooks:
       - id: check-added-large-files
+      - id: no-commit-to-branch
+        args:
+          # We include branch names in various AWS resource names, including S3 buckets, so:
+          #   * They must not be too long.
+          #   * They must contain only lowercase letters, numbers, hyphens, and dots (and dots are not recommended).
+          #   * See https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html
+          #   * This check is not exhaustive, but it should catch most deploy problems early.
+          - "--pattern"
+          - "^.{30,}$"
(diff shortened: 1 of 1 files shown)


gmrabian (approved):
😎

ntsummers1 (approved):
🌟

jessabean (approved):

PR #596, opened 2024-05-13
Add helpers for creating/updating reports and template versions

benmartin-coforma:
### Description
This is a continuation of the work in PR #582. There shouldn't be any surprises here; we're just filling out the rest of the helper methods needed for reports. Once it is merged, there will be no direct DynamoDB or S3 access in any of the report handlers.

**Detailed list of moderately-interesting changes**
* create.test - some tests were actually testing "deeply", verifying functionality in copy.ts or formTemplates.ts. Those functions are now mocked, simplifying the tests here. The unit tests in copy.test.s, formTemplates.test.ts, and transformations.test.ts already cover their own functionality.
* create.ts - remove some deprecated parameters when calling helper methods
* update.ts - remove a check that would catch a metadata object without formTemplateId or fieldDataId. These fields are guaranteed to be present by the type system, so the check just ends up adding noise.
* storage/reports.ts - New methods for creating & updating metadata, field data, and form templates. Also new methods for creating & querying form template _verisons_.
   * I made a call here, to create `queryLatestFormTemplateVersionNumber` instead of `queryLatestFormTemplate`. This is an outlier in the file, since every other method deals in whole objects, but the call site in formTemplates.ts isn't interested in the rest of the template, only that number. Putting that property access & the `?? 0` logic in here means that this is the only function in the file that requires two unit tests. Was this a good choice?
* constants.ts - Moved a few constants to reports.ts, since they are only used there. Some maps have more specific type annotations, so that TS can help us more if/when the report types ever change.
* formTemplates.ts - rework formTemplateForReportType to briefer & hide more. Removed error logging S3 & dynamo interaction - it would be redundant with the logging in the AWS clients.
* copy.ts - a bit stronger typing in here, but more to do later. But that will require more code changes & deserves its own PR.
* utils/other.test.ts - 60% whitespace changes, sorry. The other 40% are for functionality previously covered in reports/create.test.ts.
* utils/other.ts - Remove `lastCreatedWorkPlan` in favor of a `.filter` and a `.reduce`. Also, `getLastCreatedWorkPlan` is significantly simplified now that it uses storage helpers.
* setupJest.ts - Remove some mock data previously used when testing report handlers
* types/other.ts - Rename `FormTemplate` interface - these objects only contain versioning information. Now we have `FormTemplateVersion` and `ReportFormTemplate` for our function names, and `FormTemplateVersion` and `ReportJson` are the corresponding type names. Maybe someday, rename the `ReportJson` interface to `ReportFormTemplate`? But not today.
* types/reportContext.ts - Explicitly add `Choice[]` to the `ReportFieldData` type. In a lot of ways we can get away with treating choice arrays as entities, but I feel it'll be clearer to make the distinction whenever we can. This sets us up to do so going forward - notably, in copy.ts.
* types/reports.ts - `ReportJson.type` is always present, and it is always a `ReportType`. Marking it as such allows `putFormTemplateVersion` to take only a single parameter; otherwise we would need to pass `reportType` through separately.

### Related ticket(s)
n/a

---
### How to test
As usual, there should be no observable behavioral changes. But since every endpoint was touched, a regression test of the entire app is appropriate.

---
### Pre-review checklist

---
#### Review

#### Security

### DIFF
--- a/services/app-api/handlers/reports/approve.test.ts
+++ b/services/app-api/handlers/reports/approve.test.ts
@@ -1,20 +1,15 @@
 import { approveReport } from "./approve";
-import { DynamoDBDocumentClient, PutCommand } from "@aws-sdk/lib-dynamodb";
-import { mockClient } from "aws-sdk-client-mock";
 // utils
 import { proxyEvent } from "../../utils/testing/proxyEvent";
 import { mockWPReport } from "../../utils/testing/setupJest";
 import { error } from "../../utils/constants/constants";
-import { getReportMetadata } from "../../storage/reports";
+import { getReportMetadata, putReportMetadata } from "../../storage/reports";
 // types
 import { APIGatewayProxyEvent, StatusCodes } from "../../utils/types";
 
-const dynamoClientMock = mockClient(DynamoDBDocumentClient);
-
 jest.mock("../../storage/reports", () => ({
-  getReportFieldData: jest.fn(),
-  getReportFormTemplate: jest.fn(),
   getReportMetadata: jest.fn(),
+  putReportMetadata: jest.fn(),
 }));
 
 jest.mock("../../utils/auth/authorization", () => ({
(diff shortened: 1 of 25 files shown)


gmrabian (approved):
deployed env looks good! I am ok with the queryByHash function going in the storage/reports file

ntsummers1 (approved):
Sheeeesh. Theres some good work here. Deployed env looks good so no worries there. Also loving the test coverage increasing with this. Nice job!

jessabean (approved):
🙏🏻  bless you

PR #605, opened 2024-05-21
Refactor AddNewWorkPlanModal into AddEditReportModal

gmrabian:
### Description

Compare to PR #616 

Refactor the two dashboard modals into one that accommodates all scenarios for WP and SAR creation.
A lot of the new logic in `AddEditReportModal.tsx` and the tests are from the deleted modal files.

`components/modals/AddEditReportModal.tsx`
- update double equals to triple throughout
- add ability to show both WP form types
- update `prepareWpPayload` and `writeReport` with logic from the deleted AddNewWorkPlanModal
- update jsx to simplify logic and use full capability of modal
- remove unused styling (sx)

`components/modals/AddEditReportModal.test.tsx`
- add test component for new wp
- add test component for editing sar
- copy tests for new work plans from `AddNewWorkPlanModal.test.tsx` into this file
- update SAR tests
- add a11y tests for all view types

`components/pages/Dashboard/DashboardPage.tsx`
- remove logic that split which modal is opened

deleted `components/modals/AddNewWorkPlanModal.test.tsx`
deleted `components/modals/AddNewWorkPlanModal.tsx`
removed from `components/index.ts`

`tests/cypress/support/commands.js`
- Add extra wait after clearing sessions to ensure we start fresh

Open to feedback! Do we like this better?
In doing this refactor I realized that maybe grouping both WP types into one component, and keeping SAR separate, could be another way to organize it. I know that would give us two components again, but maybe the logic would be simpler when drawn along those lines. What do you think? Do you think this component is logical as is?

### Related ticket(s)

CMDCT-3473

---
### How to test

Deployed url: https://d176fbxzy8k4fx.cloudfront.net/
Fastest to test locally with the postman collection:
- Log in as a state user
- Create a WP and ensure you get the first work plan form (asks for year and period)
- Use the postman collection to fill out, submit, and approve the report
- In the UI, create a copy WP and ensure the correct modal shows
- Go to SAR dashboard and ensure that modal works as expected

---
### Pre-review checklist


jessabean:
I haven't done a complete code review of this PR yet, but these are some general thoughts:

I do think I like the idea of having separate components for WP and SAR. Maybe an unpopular opinion. 😬  I like components that are as simple as possible and responsible for one thing. Looking back over the code, I see that there are a fair amount of variables and functions that are only applicable to one report type or the other; just think it might be easier to read if we split them out so we don't have to read through SAR payload stuff if we're not debugging the SAR.

I was also wondering why this component is called the `AddEdit` report modal. It doesn't seem like we are editing anything? It seems like `NewWorkPlanModal` or `CreateWorkPlanModal` might be more accurate.

services/ui-src/src/components/modals/AddEditReportModal.tsx
```diff
@@ -246,34 +286,17 @@ export const AddEditReportModal = ({
...
-  modalHeader: {
-    padding: "2rem 2rem 0 2rem",
-  },
-  modalBody: {
-    padding: "1rem 2rem 0 2rem",
-  },
-  modalFooter: {
-    justifyContent: "flex-start",
-    padding: "0 2rem 2rem 2rem",
-  },
```
jessabean, line 276: 👏🏻 <standing ovation> 👏🏻

gmrabian:
> I haven't done a complete code review of this PR yet, but these are some general thoughts:
> 
> I do think I like the idea of having separate components for WP and SAR. Maybe an unpopular opinion. 😬 I like components that are as simple as possible and responsible for one thing. Looking back over the code, I see that there are a fair amount of variables and functions that are only applicable to one report type or the other; just think it might be easier to read if we split them out so we don't have to read through SAR payload stuff if we're not debugging the SAR.
> 
> I was also wondering why this component is called the `AddEdit` report modal. It doesn't seem like we are editing anything? It seems like `NewWorkPlanModal` or `CreateWorkPlanModal` might be more accurate.

I think the AddEdit name came from MCR, and it is applicable for SAR because you can edit the report via the modal after creation.
I am open to trying to split them! Maybe I will make another branch and we can compare 🙂

gmrabian:
closing. we selected https://github.com/Enterprise-CMCS/macpro-mdct-mfp/pull/616

PR #608, opened 2024-05-21
Cloudfront Policy Headers

BearHanded:
### Description

Switch from a Cloudfront Function to Policies to apply security headers. The function is not applied on a page not found, which is how our SPA magic works. A route like `mysite.com/home` would not get the function, but `mysite.com` would, despite them both returning the same item from S3 at mysite.com's index file.

### Related ticket(s)

N/A

---
### How to test

Deployed instance should still have the security headers, and nested pages should have them as well, when inspecting hte network tab headers.
Log in, go to a page with a nested path. Refresh. Same headers should be contained.

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/ui/serverless.yml
+++ b/services/ui/serverless.yml
@@ -235,9 +235,7 @@ resources:
               Cookies:
                 Forward: none
             ViewerProtocolPolicy: redirect-to-https
-            FunctionAssociations:
-              - EventType: viewer-response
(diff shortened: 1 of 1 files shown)


ailZhou (approved):
tested deploy, header exist when refreshing in a nested path.

PR #625, opened 2024-06-05
[DO NOT MERGE] Playwright Demo

jessabean:
### Related ticket(s)

CMDCT-

---
### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---


jessabean:
Comparing a couple of tests, speed-wise (even if it's not 100% parity in terms of locators)

|   |  Cypress  |  Playwright  |
| -- | -- | -- |
| **login** |  12s | 2.5s |
| **home** | 9s | 2.5s |

karla-vm:
I don't actually mind that Playwright is more verbose than Cypress if it's even more readable / understandable for the developer. The fact that we only have one dependency (versus 42 with Cypress) would make our codebase a bit lighter too, so that's a big plus! I like it.

package.json
```diff
@@ -10,7 +10,9 @@
   "scripts": {
     "db:seed": "cd services/app-api && yarn run db:seed",
     "test": "cd tests/cypress && npm test && cd -",
-    "test:ci": "cd tests/cypress && yarn run test:ci"
+    "test:ci": "cd tests/cypress && yarn run test:ci",
+    "test:e2e": "yarn playwright test",
+    "test:e2e-ui": "yarn playwright test --ui"
```
gmrabian, line 15: Remove redundant lead `yarn` command
```suggestion
    "test:e2e": "playwright test",
    "test:e2e-ui": "playwright test --ui"
```

playwright.config.ts
```diff
@@ -0,0 +1,59 @@
...
+  /* Retry on CI only */
+  retries: process.env.CI ? 2 : 0,
+  /* Opt out of parallel tests on CI. */
+  workers: process.env.CI ? 1 : undefined,
+  /* Reporter to use. See https://playwright.dev/docs/test-reporters */
+  reporter: "html",
+  /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
+  use: {
+    /* Base URL to use in actions like `await page.goto('/')`. */
+    baseURL: "http:localhost:3000",
```
gmrabian, line 29:
```suggestion
    baseURL: "http://localhost:3000",
```

playwright.config.ts
```diff
@@ -0,0 +1,59 @@
...
+    {
+      name: "webkit",
+      use: { ...devices["Desktop Safari"] },
+    },
+  ],
+
+  /* Run your local dev server before starting the tests */
+  webServer: {
+    command: "./run local --update-env",
+    url: "http:localhost:3000",
```
gmrabian, line 56:
```suggestion
    url: "http://localhost:3000",
```

playwright.config.ts
```diff
@@ -0,0 +1,59 @@
...
+    reuseExistingServer: !process.env.CI,
+  },
```
gmrabian, line 58: I tried to add another config option here `stdout: 'pipe',` so we could see the app running, but it didn't print to terminal. Not sure if there's other things we can try, but it'd be nice for the terminal logs to appear

playwright.config.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    {
+      name: "firefox",
+      use: { ...devices["Desktop Firefox"] },
+    },
+
+    {
+      name: "webkit",
+      use: { ...devices["Desktop Safari"] },
+    },
```
gmrabian, line 50: I'd maybe remove these since we are only using chrome. Fewer installs from playwright

tests/e2e/login.spec.ts
```diff
@@ -0,0 +1,41 @@
+import { test, expect } from "@playwright/test";
```
gmrabian, line 1: for these two login tests could we not re-use the helper logic for the login?

gmrabian (commented):
[Mako reference playwright config](https://github.com/Enterprise-CMCS/macpro-mako/blob/master/src/services/ui/playwright.config.ts)
[Mako reference e2e github workflow step](https://github.com/Enterprise-CMCS/macpro-mako/blob/master/.github/workflows/deploy.yml#L124-L152)

PR #649, opened 2024-06-13
Cmdct 3662 - remove button from inside link

britt-mo:
### Description
The "Leave form" call to action should be marked up as a link and there should be no <button> element inside.
The "Leave form" call to action should be styled as an outline button

original:

Screenshot 2024-06-13 at 9 53 17 AM

### Related ticket(s)

CMDCT-3662

---
### How to test
Login, create WP, enter WP and see that 'Leave form' still looks the same but is no longer a `button`.

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/ui-src/src/components/layout/Header.tsx
+++ b/services/ui-src/src/components/layout/Header.tsx
@@ -177,7 +177,7 @@ const sx = {
   },
   leaveFormText: {
     border: "1px solid",
-    padding: ".5rem 1.5rem",
+    padding: ".5rem 1rem",
     borderRadius: "5px",
     color: "palette.primary",
     fontWeight: "bold",
--- a/services/ui-src/src/components/layout/Header.tsx
+++ b/services/ui-src/src/components/layout/Header.tsx
@@ -72,11 +72,11 @@ export const Header = ({ handleLogout }: Props) => {
                   as={RouterLink}
                   to={report?.formTemplate.basePath || "/"}
                   sx={sx.leaveFormLink}
-                  variant="unstyled"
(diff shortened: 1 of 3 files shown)


gmrabian (approved):
do we need this in MCR as well?

britt-mo:
> do we need this in MCR as well?

@gmrabian yeah, I can see if there's already a card for it, if not I'll make one

services/ui-src/src/components/layout/Header.tsx
```diff
@@ -183,4 +175,11 @@ const sx = {
   closeIcon: {
     width: "2rem",
   },
+  leaveFormText: {
```
jessabean, line 178: Idea: what do you think of moving all of these styles into a new variant for the link in the `theme.ts` file? So instead of using the `unstyled` variant, we could add a new one like `unstyledButton` (or whatever sounds good).

```js
Link: {
  variants: {
  // all the other variants would be above... 
    unstyledButton: {
          color: "palette.primary",
          border: "1px solid",
          padding: ".5rem 1rem",
          borderRadius: "5px",
          fontWeight: "bold",
          textDecoration: "none",
          _visited: {
            color: "palette.primary",
          },
          ":hover, :visited:hover": {
            color: "palette.primary_darker",
            textDecoration: "none",
          },
        },

  }
}
```

Alternatively, we could move this block up to the `leaveFormLink` property but might have to do some additional tinkering to override some design system component styling

jessabean (approved):

karla-vm (approved):

PR #659, opened 2024-06-18
Use concurrency to manage action queueing

gmrabian:
### Description

Remove script that manually locked branches from triggering multiple deletions. Deleting a branch will only trigger the action once, and if it is manually triggered more times in a short period then failure is expected.

Also explicitly declaring serverless should install v3 until we decide our migration path

### Related ticket(s)

CMDCT-3642

---
### How to test

No significant change to destroy action after this merges

---
### Pre-review checklist


karla-vm (approved):
sure!

PR #666, opened 2024-06-24
Refactor Entity Status Icon

gmrabian:
### Description

Refactor the EntityStatusIcon component to
- Simplify logic
- Remove duplicate styles
- Unify icons in all views
- Make the status types fit in the enum

Notes for review: I recommend you review the commits one-by-one

### Related ticket(s)

CMDCT-3690

---
### How to test

- Go to https://d11w0gb4ih2yvu.cloudfront.net/
- Log in as `cypressstateuser@test.com`
- View the initiative sections for WP and SAR, in both report and PDF views
  - Verify the icons are consistent
- Modify one of the initiatives in the SAR
  - Verify the statusing works as expected

No changes to display or logic (other than using lighter version of checks and exes for all locations)
---
### Pre-review checklist


services/ui-src/src/components/tables/getEntityStatus.tsx
```diff
@@ -84,12 +84,16 @@ export const getEntityStatus = (
     });
 
     //check to see if each validation id was matched to user selected values
-    return isFilled?.every((field: AnyObject) => {
-      return field && field.length > 0;
-    });
+    if (
+      isFilled?.every((field: AnyObject) => {
+        return field && field.length > 0;
```
bangbay-bluetiger, line 89: `field?.length > 0` ?

gmrabian: I don't think we need the optional chain if it's combined with `field &&` because the first part ensures it exists

gmrabian: Hmm maybe not 🤔 

Screenshot 2024-06-25 at 12 19 02 PM

bangbay-bluetiger: Looks like `var name = undefined` actually assigns a string `undefined` so that's not correct behavior.

services/ui-src/src/components/tables/EntityStatusIcon.tsx
```diff
@@ -1,65 +1,53 @@
...
+      case EntityStatuses.INCOMPLETE:
         return {
-          src: isPdf ? unfinishedIconDark : unfinishedIcon,
-          alt: isPdf ? "" : "warning icon",
+          src: unfinishedIcon,
+          alt: showLabel ? "" : "warning icon",
           style: sx.errorText,
           text: "Error",
         };
+      case EntityStatuses.NO_STATUS:
```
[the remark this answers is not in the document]
jessabean: Agreed; I think we can just let default handle these cases.

services/ui-src/src/components/tables/getEntityStatus.tsx
bangbay-bluetiger: Also `?` for `isFilled?` seems unnecessary because `isFilled` is always an array.

services/ui-src/src/components/tables/EntityStatusIcon.tsx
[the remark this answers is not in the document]
gmrabian: I do think it's nice to define these, given it's an enum, just so that it's clear to other developers those situations are acknowledged and accounted for in the switch statement, even if they are ignored.
I'm ok with removing them if you or others feel strongly about it though

bangbay-bluetiger (approved):

services/ui-src/src/components/tables/getEntityStatus.tsx
gmrabian: hmm that's a good point. I'm gonna leave that alone for now because I saw a lot more opportunities for refactoring here

britt-mo (approved):
Reviewed per commit, thanks for the suggestion. Walked through with the cloudfront site and everything appeared as expected.

jessabean (approved):

PR #672, opened 2024-06-26
Adding continue on failure option for flaky playwright tests 

dwhitestratiform:
### Description
Playwright tests are still a WIP and we're seeing some failures in main and are still working through why that is. Rather than revert or remove tests this option allows us to keep the tests in place, monitor failures but wont fail the job for a failing playwright test.

we should however remove this option when/if we make the decision to move to playwright full time and rip out cypress. 

### Related ticket(s)
n/a

---
### How to test
see failed tests but the job in github actions will be green 

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -297,6 +297,7 @@ jobs:
         run: yarn playwright install --with-deps
       - name: Run Playwright tests
         run: yarn playwright test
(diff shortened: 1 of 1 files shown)


jessabean (approved):
oh sure, why not? 😆

PR #679, opened 2024-06-28
Organize theme files

jessabean:
### Description
[RFC](https://docs.google.com/document/d/1zyHGfboYyhncV0MXnoCNipn1xWoD8IfAS_oVK5-FExk/edit#heading=h.rz1wnztilpui)

Proposed organization of theme files. Breaking up main theme file into more atomic pieces that include foundation styles and component styles, along with their variants.

### To Test
- This kind of touches every part of the app (ugh, sorry), so it will require looking around at all the pages
- Everything should look mostly the same and nothing obviously broken

### DIFF
--- a/services/ui-src/src/styles/components/link.ts
+++ b/services/ui-src/src/styles/components/link.ts
@@ -5,7 +5,7 @@ const baseStyles = {
   transition: "all 0.3s ease",
 };
 
-const primaryVarian = {
+const primaryVariant = {
   color: "palette.primary",
   _visited: {
     color: "palette.primary",
--- a/services/ui-src/src/styles/components/link.ts
+++ b/services/ui-src/src/styles/components/link.ts
@@ -1,0 +1,71 @@
+import { ComponentStyleConfig } from "@chakra-ui/react";
+
+const baseStyles = {
+  textDecoration: "underline",
+  transition: "all 0.3s ease",
+};
+
+const primaryVarian = {
+  color: "palette.primary",
+  _visited: {
+    color: "palette.primary",
+    textDecorationColor: "palette.primary",
+  },
+  ":hover, :visited:hover": {
+    color: "palette.primary_darker",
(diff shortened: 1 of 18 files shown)


services/ui-src/src/styles/components/link.ts
```diff
@@ -0,0 +1,71 @@
+import { ComponentStyleConfig } from "@chakra-ui/react";
+
+const baseStyles = {
+  textDecoration: "underline",
+  transition: "all 0.3s ease",
+};
+
+const primaryVarian = {
```
gmrabian, line 8: typo
```suggestion
const primaryVariant = {
```

gmrabian (commented):
I know this is still a draft but I think it's really cool

britt-mo (commented):
disabled edit button styling 
Screenshot 2024-08-27 at 1 40 42 PM


Other than that, there are just smaller things that differ from val--buttons are a bit wider, `x Close` button in modals is positioned lower, card header text in this PR is bigger

gmrabian:
> disabled edit button styling Screenshot 2024-08-27 at 1 40 42 PM
> 
> Other than that, there are just smaller things that differ from val--buttons are a bit wider, `x Close` button in modals is positioned lower, card header text in this PR is bigger

+1

Also links remain in the unclicked styling on both this and dev so we are keeping functionality, but is that intentional?

jessabean:
> > disabled edit button styling Screenshot 2024-08-27 at 1 40 42 PM
> > Other than that, there are just smaller things that differ from val--buttons are a bit wider, `x Close` button in modals is positioned lower, card header text in this PR is bigger
> 
> +1
> 
> Also links remain in the unclicked styling on both this and dev so we are keeping functionality, but is that intentional?

Garrett, I'm not sure I understand this question! Can you show me an example? It's possibly a regression, but I think I need an example to be sure.

karla-vm (approved):

PR #680, opened 2024-06-28
refactor EntityCardBottomSection cmdct-3359

britt-mo:
### Description
Refactoring effort around EntityCard functionality

General changes:
- The contents of each entity has been separated into their own components
- Only `ObjectiveProgress` entity still uses `completedEntity` prop, the other two entities only use `formattedEntityData`
- No more `Top Section` or `Bottom Section`
- `EntityCard` has been renamed to `EntityStepCard`
- `ExportedEntityStepCard` was created to replace the use of the `printVersion` prop

There are still other improvements that Jess suggested around the way we display the text content along with the styling of the cards. I figured this effort could be focused on what I've outlined above, and then we could have another card that is focused on the other refactoring suggestions.

I tried to have descriptive commit messages so it may be helpful to look at the code changes in blocks per specific commits.

### Related ticket(s)

CMDCT-3359

---
### How to test
Create a WP and SAR and check that in all states (incomplete, complete, and export) there is no change to the UI.

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---


services/ui-src/src/components/cards/ObjectiveProgressCardBottomSection.tsx
```diff
@@ -55,107 +55,93 @@ const wereTargetsMetForObjectiveProgress = (
   );
 };
 
-export const EntityStepCardBottomSection = ({
-  stepType,
+export const ObjectiveProgressCardBottomSection = ({
```
bangbay-bluetiger, line 58: Find the `ObjectiveProgressCardBottomSection` naming a little strange since nothing else is named `ObjectiveProgressCard`. It makes me wonder where are the other parts of this card?

services/ui-src/src/components/cards/EntityCardByStepType.tsx
```diff
@@ -9,12 +9,16 @@ import {
   ReportType,
 } from "types";
 import { useStore } from "utils";
+import { ObjectiveProgressCardBottomSection } from "./ObjectiveProgressCardBottomSection";
 
-export const EntityStepCardTopSection = ({
+export const EntityCardByStepType = ({
```
bangbay-bluetiger, line 14: This file seems a bit unruly. I'd prefer each step component to pulled put into its own file since there's no shared HTML between them.

bangbay-bluetiger (changes requested):

jessabean:
Okay my thoughts on the whole refactoring effort are kind of big (and all of this has less to do with your approach and more to do with the definition of the work to do in the JIRA ticket). As with all of the other stuff in our app, this humble little component is really, really complex!

I went through all the variations of EntityStepCard and learned some things, much of which is documented in this [Mural board](https://app.mural.co/t/coforma8350/m/coforma8350/1719925014903/87008cd0cd5771820f0ef1a52468d81d74ce2690?sender=ubabab5dfd4cc79dd449f6665).

1. At the end of the day, it would be nice to not have so much template logic to render different giant blocks of HTML for each step type. That `switch` is hella unruly, like Bangbay pointed out.
2. The majority of the content for each step type is structured in the exact same way: label + description. There are two exceptions:
    - quarterly content, which has a label + a nested grid of label/content for each quarter
    - progress content, which has a colored background but is really just another label + description styled in a different way
3. Because the progress content is basically the same as all the other label + description content, I don't think it really deserves to be a separate component, whether it is called `EntityStepCardBottomSection` or `ObjectiveProgressBottomSection`. It's just another `Box` that has a variant to determine background color.
4. For consistency's sake with other PDF-related components, I might consider creating a separate component for the PDF version of this card, if only to get all the `printVersion` logic outta here.

**Okay Jess WTF that is a lot what should we do?**

I think this is a refactor that could happen in stages. And maybe the very first step is to leave this component better than we found it. Simplify it a teeny bit. And next time, we simplify it a teeny bit more.

Some suggested teeny steps to move forward:

- Split out the `printVersion` content into something like `ExportedEntityStepCard` and drop it into the `/export` folder. End state: there is no need for a `printVersion` prop in the `EntityStepCard` or section components, and we render the exported version in the PDF instead of sharing this one.
- Split out the `EntityCardByStepType` content into 3 separate mini components, one for each step type (for now...ultimately I think this is still kinda messy but hopefully it is temporary). Then, move the logic for which component to render into `EntityStepCard`. 
    ```js
      // EntityStepCard.tsx
     const cardContent = () => {
        switch (stepType) {
            case OverlayModalStepTypes.OBJECTIVE_PROGRESS:
                return <EntityStepObjectiveProgress />;
            case OverlayModalStepTypes.EVALUATION_PLAN:
                return <EntityStepEvaluationPlan />;
            case OverlayModalStepTypes.FUNDING_SOURCES:
                return <EntityStepFundingSources />;
            default:
                return null;
        }
    };


   // line 172 of EntityStepCard becomes:
  { cardContent() }
    ```
- 😈  Refactor the styling so we are using a nice clean grid instead of absolutely positioning the icons and actions. MWAHAHAHA.

services/ui-src/src/components/cards/EntityStepCard.test.tsx
```diff
@@ -175,12 +154,11 @@ const NoOpenAddEditEntityFunction = (
...
     openDeleteEntityModal={openDeleteEntityModal}
     openDrawer={mockOpenDrawer}
-    printVersion={false}
   />
 );
 
-describe("<EntityCard />", () => {
-  describe("Completed EntityCard", () => {
+describe("<EntityStepCard />", () => {
+  describe("Completed EntityStepCard", () => {
```
bangbay-bluetiger, line 161: I'm seeing `Warning: React does not recognize the entityIndex prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase entityindex instead. If you accidentally passed it from a parent component, remove it from the DOM element.` from running tests.

services/ui-src/src/components/export/ExportedOverlayModalReportSection.tsx
```diff
@@ -56,15 +56,14 @@ export const ExportedOverlayModalReportSection = ({
         const nextHeadingLevel = `h${nextLevel}`;
 
         return (
-          <EntityStepCard
+          <ExportedEntityStepCard
```
bangbay-bluetiger, line 59: Tests for this component now have `Warning: React does not recognize the headingLevel prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase headinglevel instead. If you accidentally passed it from a parent component, remove it from the DOM element.`

services/ui-src/src/components/cards/EntityStepCard.tsx
```diff
@@ -20,31 +16,39 @@ import deleteIcon from "assets/icons/icon_cancel_x_circle.png";
...
   hasBorder,
-  headingLevel,
   ...props
 }: Props) => {
   let entityCompleted = false;
-  const entitiesCount = `${entityIndex + 1} / ${entityTotal}`;
   const { report } = useStore() ?? {};
   // any drawer-based field will do for this check
+
+  let cardContent = {};
```
bangbay-bluetiger, line 39: Could have it as `let cardContent = <></>;` so we know it's always a component.

services/ui-src/src/components/cards/EvaluationPlanEntity.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+const evaluationPlanEntity = (
+  <EvaluationPlanEntity
+    formattedEntityData={mockEvaluationPlanFormattedEntityData}
+  />
+);
+
+describe("<EvaluationPlanEntity />", () => {
+  beforeEach(() => {
+    render(evaluationPlanEntity);
```
bangbay-bluetiger, line 14: Having `render(evaluationPlanEntity);` here means it'll double render in `testA11y` which may not be a problem here but causes problems in some other tests. If you have only one test, you don't really need the `beforeEach` block and can do this instead:
```
test("EvaluationPlanEntity renders correctly", () => {
  render(evaluationPlanEntity);
  expect(
    screen.getByRole("heading", { name: "mockEvaluationPlanName" })
  ).toBeVisible();
});
```

services/ui-src/src/components/cards/EvaluationPlanEntity.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+  });
+  afterEach(() => {
+    jest.clearAllMocks();
```
bangbay-bluetiger, line 17: There isn't anything you need to clear here, so I'd take out this block.

services/ui-src/src/components/cards/EvaluationPlanEntity.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+  });
+
+  test("EvaluationPlanEntity renders correctly", () => {
+    expect(
```
bangbay-bluetiger, line 21: There are some other scenarios that would be good to test. I'll note in the component file.

services/ui-src/src/components/cards/EvaluationPlanEntity.tsx
```diff
@@ -0,0 +1,131 @@
...
+        {formattedEntityData.objectiveName}
+      </Heading>
+      <Text sx={sx.subtitle}>
+        Performance measure description or indicators your state or territory
+        will use to monitor progress towards achievement
+      </Text>
+      <Text sx={sx.description}>{formattedEntityData.description}</Text>
+      <Text sx={sx.subtitle}>Performance measure targets</Text>
+      <Text sx={sx.description}>{formattedEntityData.targets}</Text>
+      {report?.reportType === ReportType.WP && (
```
bangbay-bluetiger, line 25: This would be helpful to have unit tests for.

services/ui-src/src/components/cards/EvaluationPlanEntity.tsx
```diff
@@ -0,0 +1,131 @@
...
+        <>
+          <Text sx={sx.subtitle}>
+            Does the performance measure include quantitative targets?
+          </Text>
+          <Text sx={sx.description}>
+            {formattedEntityData?.includesTargets}
+          </Text>
+        </>
+      )}
+      {formattedEntityData.quarters?.length > 0 && (
```
bangbay-bluetiger, line 35: This would be helpful to have unit tests for.

services/ui-src/src/components/cards/FundingSourcesEntity.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+const fundingSourcesEntity = (
+  <FundingSourcesEntity
+    formattedEntityData={mockFundingSourcesFormattedEntityData}
+  />
+);
+
+describe("<FundingSourcesEntity />", () => {
+  beforeEach(() => {
+    render(fundingSourcesEntity);
```
bangbay-bluetiger, line 14: Similar comments from `EvaluationPlanEntity.test.tsx`.

services/ui-src/src/components/cards/FundingSourcesEntity.tsx
```diff
@@ -0,0 +1,87 @@
...
+export const FundingSourcesEntity = ({
+  formattedEntityData,
+  headingLevel = "h2",
+}: Props) => {
+  return (
+    <>
+      <Heading as={headingLevel} sx={sx.mainHeading}>
+        {formattedEntityData.fundingSource}
+      </Heading>
+      {formattedEntityData.quarters.length > 0 && (
```
bangbay-bluetiger, line 16: This would be helpful to test.

services/ui-src/src/components/cards/ObjectiveProgressEntity.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+const objectiveProgressEntityComponent = (
+  <ObjectiveProgressEntity
+    formattedEntityData={mockObjectiveProgressFormattedEntityData}
+    entityCompleted={true}
+  />
+);
+
+describe("<ObjectiveProgressEntity />", () => {
+  beforeEach(async () => {
```
bangbay-bluetiger, line 21: `async` seems unnecessary. There's no `await` being used.

services/ui-src/src/components/cards/EvaluationPlanEntity.tsx
```diff
@@ -0,0 +1,131 @@
...
+import { notAnsweredText } from "../../constants";
+// utils
+import { AnyObject, HeadingLevel, ReportType } from "types";
+import { useStore } from "utils";
+
+export const EvaluationPlanEntity = ({
+  formattedEntityData,
+  headingLevel = "h2",
+}: Props) => {
+  const { report } = useStore() ?? {};
```
bangbay-bluetiger, line 12: Looks like `const { report } = useStore();` is fine because `report?` is used when needed.

services/ui-src/src/components/cards/FundingSourcesEntity.tsx
```diff
@@ -0,0 +1,87 @@
...
+        <>
+          <Text sx={sx.subtitle}>Projected quarterly expenditures</Text>
+          <Grid sx={sx.grid}>
+            {formattedEntityData?.quarters?.map((quarter: any) => {
```
bangbay-bluetiger, line 20: `?` checks aren't needed here.

services/ui-src/src/components/cards/ObjectiveProgressEntity.tsx
```diff
@@ -55,17 +54,45 @@ const wereTargetsMetForObjectiveProgress = (
...
+      <Text sx={sx.description}>{formattedEntityData.targets}</Text>
+      {formattedEntityData.quarterProjections.length > 0 && !entityCompleted && (
         <>
-          {(formattedEntityData?.performanceMeasureProgress ||
-            printVersion) && (
+          <Text sx={sx.subtitle}>
+            Quantitative targets for this reporting period
+          </Text>
+          <Grid sx={sx.sarGrid}>
+            {formattedEntityData?.quarterProjections.map((quarter: any) => {
```
bangbay-bluetiger, line 80: `?` not needed on `formattedEntityData`.

services/ui-src/src/components/cards/ObjectiveProgressEntity.tsx
```diff
@@ -55,17 +54,45 @@ const wereTargetsMetForObjectiveProgress = (
...
+                  </Flex>
+                </GridItem>
+              );
+            })}
+          </Grid>
+        </>
+      )}
+      {entityCompleted ? (
+        <>
+          {formattedEntityData?.performanceMeasureProgress && (
```
bangbay-bluetiger, line 95: `formattedEntityData` is always defined so `formattedEntityData?` shouldn't be needed anywhere in this file.

services/ui-src/src/components/export/ExportedEntityStepCard.tsx
```diff
@@ -0,0 +1,200 @@
...
+  hasBorder,
+  ...props
+}: Props) => {
+  let entityCompleted = false;
+  const entitiesCount = `${entityIndex + 1} / ${entityTotal}`;
+  // any drawer-based field will do for this check
+  let cardContent = {};
+  switch (stepType) {
+    case OverlayModalStepTypes.OBJECTIVE_PROGRESS:
+      entityCompleted = formattedEntityData?.objectiveName;
```
bangbay-bluetiger, line 34: `formattedEntityData` is always defined so can remove `?`s with it in this file.

bangbay-bluetiger (commented):
Lots of improvement happening here!

services/ui-src/src/components/cards/EvaluationPlanEntity.test.tsx
britt-mo: Gotcha. I do plan to add more tests to these entity components so I'll be making updates.

services/ui-src/src/components/cards/FundingSourcesEntity.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+);
+
+const fundingSourcesEntityNoQuarters = (
+  <FundingSourcesEntity
+    formattedEntityData={mockFundingSourcesFormattedEntityDataNoQuarters}
+  />
+);
+
+describe("<FundingSourcesEntity />", () => {
+  test("FundingSourcesEntity renders correctly", () => {
```
bangbay-bluetiger, line 22: I'd update the descriptions for the tests to be
`renders with quarters` and `renders without quarters`

Screenshot 2024-07-10 at 2 39 31 PM

services/ui-src/src/components/cards/ObjectiveProgressEntity.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+    ).toBeVisible();
+    expect(
+      screen.getByText(mockObjectiveProgressFormattedEntityData.targetsMet)
+    ).toBeVisible();
+    expect(
+      screen.getByText(
+        mockObjectiveProgressFormattedEntityData.missedTargetReason
+      )
+    ).toBeVisible();
+  });
```
bangbay-bluetiger, line 57: This can one test block with `render(objectiveProgressEntityComponent);` inside it. (So it's isolated from `testA11y`.)

You could also simplify this with a loop, something like:
```
[ text1, text2, text3 ].forEach((text) => expect(screen.getByText(text)).toBeVisible())
```

bangbay-bluetiger (approved):
Non-blocking comments.

bangbay-bluetiger (approved):

karla-vm (approved):
great work on this, Britt 🥇

PR #691, opened 2024-07-10
Only run upload report in lower envs

gmrabian:
### Description

I noticed an issue where Upload Report for playwright was running and failing in production even though the tests do not run in production. I think it was because of the `if: always()` condition. I modified it to be "always so long as it's not production"

[Production deploy with failed upload](https://github.com/Enterprise-CMCS/macpro-mdct-mfp/actions/runs/9877736178/job/27280595541)

### Related ticket(s)

CMDCT-

---
### How to test

- Next time prod deploys it does not fail this step, or run it at all

---
### Pre-review checklist

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -317,10 +317,10 @@ jobs:
           retention-days: 30
 
   upload-reports:
-    name: Uppload Reports
+    name: Upload Reports
     needs:
       - test
-    if: always()
+    if: ${{ always() && github.ref_name != 'production' }}
     runs-on: ubuntu-latest
     outputs:
       timestamp: ${{ steps.timestampid.outputs.timestamp }}


britt-mo (approved):

.github/workflows/deploy.yml
```diff
@@ -317,10 +317,10 @@ jobs:
           retention-days: 30
 
   upload-reports:
-    name: Uppload Reports
+    name: Upload Reports
     needs:
       - test
-    if: always()
+    if: ${{ always() && github.ref_name != 'production' }}
```
benmartin-coforma, line 323: Do we even need the `always()` anymore?

benmartin-coforma (approved):

.github/workflows/deploy.yml
gmrabian: 🤔 good question. I believe the purpose of the `always()` is to run this step even if previous steps fail. otherwise default behavior is to abandon all steps after a failure

benmartin-coforma: Makes sense! Thanks y'all

PR #730, opened 2024-08-12
cmdct-3881 - changes to archive, remove unarchive

britt-mo:
### Description
AC:

Please remove the archive/unarchive functionality from the SAR. 
For the WP:

- Remove the unarchive functionality completely. 

- Archive should only be available if the WP has not been connected to a SAR. If a WP is ineligible for Archive then the column should be blank. 

- Workplans that have been archived should have a disabled button that says "Archived"

- There should be a warning modal when an admin is archiving a report that tells them the step is irreversible (since we removed the unarchive feature). 

```
new modal: 
heading: `Are you sure you want to archive this MFP Work Plan?`

body text: `This action can’t be undone. Once the MFP Work Plan is archived, this MFP Work Plan can’t be unarchived or accessed by the state or territory. `

The user needs to enter `Archive` into the text box before they can archive. This will be the same functionality as the "Approve" modal for the WP.  
```

There is also updated copy in the instructions section of the admin view. Placement of copy for this is in magenta in Figma. 

```
 To archive a submission and hide it from a state or territory’s dashboard, use “Archive”. You can only archive an MFP Work Plan that is not associated with a MFP Semi-Annual Progress Report and once archived it can’t be unarchived.
```

### Related ticket(s)

CMDCT-3881

---
### How to test
Three areas to test:
1. WP and SAR: No option to `unarchive`
2. SAR: No option to `archive`
3. WP that does not have an associated SAR: `should` be able to be archived 
4. WP that has an associated SAR: `should not` be able to be archived

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---


karla-vm (approved):
Tested locally and everything is working as expected 💯

gmrabian (commented):
There is a playwright helper that may need to be updated too `archiveExistingWPs` in `tests/e2e/helpers.ts`

We may want to consider updating the api layer as well. Currently the logic for archiving is `archived: !currentArchivedStatus`. If someone calls the api directly they could still unarchive a report. And similarly there should be protections on the backend to ensure it's a WP and there is no associated SAR.

The disabled button just says "Archive" but the AC says "Archived" (will require test changes if we go with "Archived")
Screenshot 2024-08-13 at 2 42 50 PM

tests/e2e/helpers.ts
```diff
@@ -45,7 +45,6 @@ async function archiveReports(buttons: Locator, page) {
         response.status() == 200
     );
 
-    await archiveButtons[0].click();
     await archivePromise;
     await archiveReports(buttons, page);
   }
```
gmrabian, line 50: It looks like you lost the changes in this function when you resolved the merge conflict https://github.com/Enterprise-CMCS/macpro-mdct-mfp/pull/730/commits/0d79db3be162032c034b7383358495388a360d2c

tests/e2e/helpers.ts
```diff
@@ -61,12 +60,18 @@ export async function archiveExistingWPs({ page }) {
...
+
   await page.getByRole("button", { name: "Go to Report Dashboard" }).click();
+
   const reportsPromise = page.waitForResponse(
     (response) =>
       response.url().includes("/reports/WP/PR") && response.status() == 200
   );
   await reportsPromise;
-  await archiveReports(page.getByRole("button", { name: "Archive" }), page);
+  await page.waitForTimeout(1000);
```
gmrabian, line 71: we finally had to add a timeout, huh? Was it not loading reliably?

gmrabian:
Will we have problems with e2e tests running or rerunning as a result of this? I'm just thinking if we are unable to archive WPs with SARs and unable to archive SARs then some of our create tests will fail because we block users from creating multiple until one is done, and we have no way of getting rid of past ones

tests/e2e/helpers.ts
britt-mo: Yeah, the check for an archive button was happening before the dashboard loaded. But maybe with some re-arranging this will get fixed.

tests/e2e/helpers.ts
britt-mo: @gmrabian talked to Bangbay and I'm going to wait until her e2e work is merged in, use that +my steps for the new archiving process.

tests/e2e/wp/create.spec.ts
```diff
@@ -18,30 +19,57 @@ test("State user can create a Work Plan", async ({ page }) => {
...
+
+    const row = page.getByRole("row", { name: "Not Started" });
+    await expect(
+      row.getByRole("gridcell", {
+        name: `${stateName} MFP Work Plan ${currentYear} - Period ${secondPeriod}`,
+      })
+    ).toBeVisible();
+
+    const editButton = row.getByRole("button", { name: "Edit", exact: true });
+    await expect(editButton).toBeVisible();
```
gmrabian, line 73: nit: all of this code could be moved out of both if/else statements since it's duplicated. this might help clarify what the if/else is meant for (making sure we interact with/create the expected report)

gmrabian (approved):
Good stuff! I ran the wp and sar tests locally a bunch of times and they succeeded every time

jessabean (approved):
Nice. Everything is working as expected here 🎉

PR #740, opened 2024-08-23
archived program should be text, not disabled button

britt-mo:
### Description
Feedback from Kim regarding the changes to `archive`. After archiving, the dashboard should have text that says `Archived`, instead of a disabled outline button. Also missed a period in the accordion verbiage.

Also, not something that was a part of the original PR for CMDCT-3881, but the `View` button should never be disabled for admin.

### Related ticket(s)

CMDCT-3881

---
### How to test
Login as admin, seed the db to create a workplan. Archive it and see the `Archived` text in place of the `Archive` button. Make sure styling is the same in mobile view. See that the `View` button is enabled, regardless of the status.

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/ui-src/src/components/pages/Dashboard/DashboardPage.test.tsx
+++ b/services/ui-src/src/components/pages/Dashboard/DashboardPage.test.tsx
@@ -224,7 +224,7 @@ describe("<DashboardPage />", () => {
     });
 
     test("Clicking 'Unlock' button opens the unlock modal", async () => {
-      const unlockButton = screen.getAllByText("Unlock")[3];
+      const unlockButton = screen.getAllByRole("button", { name: "Unlock" })[3];
       expect(unlockButton).toBeEnabled();
       await userEvent.click(unlockButton);
       await expect(mockWpReportContext.releaseReport).toHaveBeenCalledTimes(1);
@@ -245,7 +245,9 @@ describe("<DashboardPage />", () => {
     });
 
     test("Clicking 'Archive' button will open the archive modal", async () => {
-      const archiveButton = screen.getAllByText("Archive")[0];
+      const archiveButton = screen.getAllByRole("button", {
+        name: "Archive",
+      })[0];
       expect(archiveButton).toBeVisible();
       await userEvent.click(archiveButton);
       await expect(
@@ -254,8 +256,8 @@ describe("<DashboardPage />", () => {
     });
 
     test("Cannot unarchive a WP", async () => {
-      const archivedText = screen.getAllByText("Archived")[1];
-      expect(archivedText).toBeInTheDocument();
+      const lastCell = screen.getAllByRole("gridcell").pop();
+      expect(lastCell).toHaveTextContent("Archived");
     });
 
     test("Can still view all WPs", async () => {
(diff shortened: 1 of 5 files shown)


jessabean (dismissed):
👍🏻

services/ui-src/src/components/pages/Dashboard/DashboardPage.test.tsx
```diff
@@ -251,10 +251,8 @@ describe("<DashboardPage />", () => {
     });
 
     test("Cannot unarchive a WP", async () => {
-      const archiveButton = screen.getAllByRole("button", {
-        name: "Archived",
-      })[0];
-      expect(archiveButton).toBeDisabled();
+      const archivedText = screen.getAllByText("Archived")[1];
```
bangbay-bluetiger, line 254: Using `getAllByText` is awkward since it'll pick up text anywhere. You could use `gridcell` in place of `button` to more closely mimic the previous test.

```
screen.getAllByRole("gridcell", {
  name: "Archived",
})[0];
```

services/ui-src/src/components/pages/Dashboard/DashboardPage.test.tsx
```diff
@@ -290,10 +288,8 @@ describe("<DashboardPage />", () => {
...
```
bangbay-bluetiger, line 291: Since the mobile and desktop tests are the same, I'm curious what would make the mobile view so different that this additional test is needed? (Just wondering for context, not suggesting any changes.)

britt-mo: I assume it's just for test coverage %

services/ui-src/src/components/pages/Dashboard/DashboardPage.test.tsx
britt-mo: but the `gridcell` one isn't the one I'm looking for, that's the `status`. The `archived` text I'm looking for is the one after it in the same row.

bangbay-bluetiger: That one also seems to be in a `gridcell`:

Screenshot 2024-08-23 at 4 07 59 PM

Maybe instead we could test that there's always two of these? Or that the last `gridcell` is the one with `Archived`?

```
const cells = screen.getAllByRole("gridcell");
const lastCell = cells[cells.length - 1];
expect(lastCell).toHaveTextContent("Archived");
```

bangbay-bluetiger: Another fancy way to write that example:
```
const lastCell = screen.getAllByRole("gridcell").pop();
expect(lastCell).toHaveTextContent("Archived");
```

britt-mo: I made that change to the desktop view, but the mobile view isn't in a table like the desktop view is.

bangbay-bluetiger (approved):

jessabean (approved):

PR #744, opened 2024-08-27
Add storage helpers for banners and templates

benmartin-coforma:
### Description
This PR wraps up the work begun in #582 and #596 , extending the storage pattern to Banners and Templates.

Calling out a few changes beyond what you may have expected based on past PRs:
* package.json: The `/storage` tests weren't being automatically run by jest; now they are!
* dynamo-lib.ts, s3-lib.ts: Since these are intended to be used only by storage helpers, rather than as general utils from anywhere in the API, I wanted to move them into the `/storage` folder.
* kafka-source-lib.ts: Everywhere _else_ in the API, we now treat S3 as a storage system for specific, strongly-typed business objects. This file is the one exception, where we treat S3 as a general JSON storage system. As such, there can be no corresponding storage helper file.
   * This is why kafka-source-lib builds its own `GetObjectCommand`
   * This also means kafka-source-lib breaks the rule I _just_ made in the previous bullet point about using s3-lib from outside `/storage`. I don't like it but I don't know what else to do.
* types/banner.ts: The Banner and AdminBanner types didn't previously exist in the API, so I copied them.
   * I don't think we ever use the types separately, so we could merge them: as far as I'm aware, every Banner in the app is an AdminBanner. But I didn't want to dive into that refactor today.
* AdminBannerProvider.tsx: Since the storage helper unwraps Dynamo's `.Item` for us, we no longer need to do so in the frontend.

### Related ticket(s)
n/a

---
### How to test
Cloudfront Environment here: https://d2quw653hu4bo1.cloudfront.net/
1. Banner tests
   1. Log in to the app as an Admin.
   2. Create a banner.
   3. Move around the app; note that the banner is visible.
   4. Delete that banner.
   5. Move around the app; note that the banner is gone.
2. Template tests
   1. Log in to the app as a state user
   2. On the home page, in the MFP Work Plan card, click the **User Guide and Help File** link
   3. Note that the file downloads successfully.
      * Note that this doesn't actually usually work in ephemeral environments. To make it work, I logged in to AWS and manually copied the help file from the `main` stack to the corresponding location in S3 for this stack.

### Notes
n/a

---
### Pre-review checklist

---
#### Review

#### Security


benmartin-coforma:
Thanks to Garrett for giving me early feedback on this PR, and helping me put it together!

gmrabian (approved):
🚀

jessabean (approved):
:shipit:

PR #746, opened 2024-09-03
Move reset button to dashboard

gmrabian:
### Description

Move the reset work plan button from the continue modal to the dashboard page
See ticket for more details

Dev choices:
- Removed reset logic from app-api (did I miss anything?) and had the frontend treat new and reset the same (both let you control the year and period)
- Gave reset modal the same form ids as create, but passed in `"clear": true` prop so that it would not try to hydrate the modal
- Changed test year to 2024 so I could test error alert

[Figma](https://www.figma.com/design/oOJMUZerbafZefLQQ33GwI/MFP_Playground?node-id=0-1&node-type=CANVAS&t=tpiSjk0ZXRgiHqiI-0)

### Related ticket(s)

CMDCT-3882

---
### How to test

https://dmprc1ij4x0jj.cloudfront.net/

I’ve set up three state users so you can see different scenarios:
1. `stateuser@test.com` existing report, not complete
2. `stateuser1@test.com` no existing reports
3. `stateuser2@test.com` existing approved report

For scenario 3
- Click “Reset MFP Work Plan”
- Select 2024 period 1
- Verify the error message pops up
- Verify it goes away if you select another year or period
- Verify you can create a reset report

If you archive the created report as an admin you can run through this flow again, or use the continue flow, etc.

### Notes

I tried adding tests for reset to `DashboardPage.test.tsx` but it caused the unlock tests to fail 🤔 
Coverage dropped on that file so if anyone has ideas let me know! I will keep trying.

---
### Pre-review checklist

---
#### Review
---

### DIFF
--- a/services/ui-src/src/components/pages/Dashboard/DashboardPage.test.tsx
+++ b/services/ui-src/src/components/pages/Dashboard/DashboardPage.test.tsx
@@ -126,7 +126,7 @@ describe("<DashboardPage />", () => {
     test("Show copied from verbiage on report versions 2 or higher", () => {
       mockedUseStore.mockReturnValue(mockUseStore);
       render(wpDashboardViewWithReports);
-      const subText = screen.getByText("copied from 2023 - Period 1");
+      const subText = screen.getByText("copied from 2024 - Period 1");
       expect(subText).toBeInTheDocument();
     });
 
--- a/services/app-api/handlers/reports/create.ts
+++ b/services/app-api/handlers/reports/create.ts
@@ -110,9 +110,6 @@ export const createReport = handler(
     const reportYear = getReportYear(reportData, overrideCopyOver);
     const reportPeriod = getReportPeriod(reportData, overrideCopyOver);
 
-    // If this Work Plan is a reset, the reporting period is the upcoming one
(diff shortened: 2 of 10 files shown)


karla-vm (approved):
Tested on the deployed instance, everything looks and works as expected! ✨

services/ui-src/src/components/pages/Dashboard/DashboardPage.test.tsx
```diff
@@ -126,7 +126,7 @@ describe("<DashboardPage />", () => {
     test("Show copied from verbiage on report versions 2 or higher", () => {
       mockedUseStore.mockReturnValue(mockUseStore);
       render(wpDashboardViewWithReports);
-      const subText = screen.getByText("copied from 2023 - Period 1");
+      const subText = screen.getByText("copied from 2024 - Period 1");
```
bangbay-bluetiger, line 129: It seems all the tests in here aren't isolated from each other, which is causing the problem that when you add a test, the other tests fail because there are elements from your new test being added to the old tests. That means that the index a lot of the other tests use to find an element (i.e. buttons[3]) has shifted. We should refactor these tests, but it's probably too big a task for this PR.

PR #750, opened 2024-09-12
Improve checks for existing report

gmrabian:
### Description

Frontend:
- Reset all component state on modal closure so that the error triggers properly on re-opening

Backend: 
- Add check to create flow that ensures unarchived report in same year and period does not exist
- Add tests

### Related ticket(s)

CMDCT-3985

---
### How to test

(if testing locally, use `yarn db:seed` to create approved WP)

https://d1vtt2axjhiobw.cloudfront.net/
- Log in as stateuser2@test.com
- Go to WP dashboard
- Open "reset" modal
- Verify alert shows as expected (when year and period match existing report)
- Close and re-open the modal multiple times. Make sure it still works as expected
- Open the network tab
- Create a reset report in a different year and period
- Resend the network request and verify it returns a 400

please log in as admin and archive any created reports to make testing easy for other users 😄

Compare to [MFP dev](https://mdctmfpdev.cms.gov/) where the erroring does not happen consistently upon reopening modal, and you can create reports in the same year and period (stateuser2 in MFP dev has an approved WP)

### Notes

---
### Pre-review checklist

---
#### Review
---

### DIFF
--- a/services/app-api/handlers/reports/create.ts
+++ b/services/app-api/handlers/reports/create.ts
@@ -20,7 +20,11 @@ import {
   getReportPeriod,
   getReportYear,
 } from "../../utils/other/other";
-import { putReportFieldData, putReportMetadata } from "../../storage/reports";
+import {
+  putReportFieldData,
+  putReportMetadata,
+  queryReportMetadatasForState,
+} from "../../storage/reports";
 import { getOrCreateFormTemplate } from "../../utils/formTemplates/formTemplates";
 import { logger } from "../../utils/debugging/debug-lib";
 import { copyFieldDataFromSource } from "../../utils/other/copy";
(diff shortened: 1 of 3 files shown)


services/app-api/handlers/reports/create.ts
```diff
@@ -196,6 +200,39 @@ export const createReport = handler(
      * to create a different SAR and attach all of its fieldData to the SAR Submissions FieldData
      */
 
+    // Validate the metadata for the submission
+    const validatedMetadata = await validateData(metadataValidationSchema, {
+      ...unvalidatedMetadata,
+    });
+
+    // Return INVALID_DATA error if metadata is not valid.
+    if (!validatedMetadata) {
```
bangbay-bluetiger, line 209: Tried to find a scenario where this would be true, but it seems that passing in undeclared fields still passes validation. Then passing in wrongly typed values causes an error to be thrown, so this should be in a try-catch block instead.

karla-vm (approved):
Functionality-wise in the deployed instance, this LGTM! I see @bangbay-bluetiger had a comment, so once that's addressed this should be good to go ✅

jessabean (approved):
I see Bangbay's comment below, but I also recognize if block exists already on main, so I think it's an improvement we can make in a a follow up.

services/app-api/handlers/reports/create.ts
gmrabian: Good catch! Although it looks like I added this it was actually elsewhere in the file and just got moved around. I made a ticket to follow up on this [CMDCT-3988](https://jiraent.cms.gov/browse/CMDCT-3988)

gmrabian:
Addressed Bangbay's comment and made a follow-up ticket.
Thanks for reviewing!

PR #752, opened 2024-09-16
FEEDBACK REQUESTED: demo new error pattern

gmrabian:
### Description

I noticed a bug where nested error throws were catching back at the handler-lib root catch.
Ex:
```js
// handler-lib.ts
try {
  create();
} catch (e) {
  console.error("handler caught");
  return response(500, e.message)
}

// create.ts
try {
  validateData(data);
} catch {
  console.error("create caught");
  return response(400, "invalid data")
}

// validateData.ts
try {
  data.validate();
} catch {
  console.error("validate caught");
  throw new Error("could not validate");
}

// OUTPUT
// validate caught
// handler caught
// Response: 500, could not validate
```

Although I am still baffled by this, it made me investigate how we are handling errors and if there is a cleaner way we can handle them.

Focus on the changes in `handler-lib.ts`, `reports/create.ts`, and `utils/constants/constants.ts`

This change shifts from determining status code alongside error to mapping error to status code and having our global handler lookup the appropriate status code.

I don't love this pattern. However, given the described bug, I am looking for a reasonable solution. Since our global handler will catch any error and default to 500, why not let it catch any error and use the correct status code too.

Also open to just fixing the bug if people know how!

### Related ticket(s)

started with CMDCT-3988 

---
### How to test

- Log in as a state user
- Open network tab and create a WP
- resend the WP request, changing a body type to be invalid (number to array `[]`, for example)
- verify it returns a 400 with the message "Provided data is not valid."

Compare to main where it returns a 500 with the same message

### Notes

A number of tests expected 500 status codes for things like invalid or missing data. This means we were accommodating the bug! We should be vigilant about creating tests that expect the _desired_ outcome, not the actual one.

---
### Pre-review checklist

---
#### Review
---


benmartin-coforma (approved):
I think this is a great idea. We would remove a lot of noise across the whole API with this approach, and I think it would be worth merging.

If you're struggling to love the errorMessage -> statusCode map, then subclassing `Error` would be an option. We could have a hierarchy like `Error <- HttpError <- BadRequestError`, and in handler-lib's `catch` we could check `if (error instanceof HttpError)` to access the status code from the specific subclass.

If you're struggling to love `throw` as control flow, then I have an alternative at PR #754. I do like the pattern I tried there, having used [something very similar in .Net Core APIs](https://learn.microsoft.com/en-us/aspnet/core/web-api/action-return-types?view=aspnetcore-8.0#actionresultt-type) in the past. We use`handlers/{resourceType}/{verb}.ts` as the layer to hold the busywork of which HTTP parameters need to be validated, who has permissions to do the HTTP action, and transforming that HTTP action into database/s3 commands; it feels appropriate to me that it would also be the layer that transforms errors into HTTP codes. But it's just an idea, and I like this PR too.

And I have no idea what's up with the nested throws! 😅 Maybe something funky with `async`?

gmrabian:
closing in favor of #754

PR #759, opened 2024-09-17
Playwright test for creating and submitting a WP

jessabean:
### Description
Adds an e2e test for creating and submitting a work plan. Utilizes the page object model to split up the various pages of the work plan, as well as dynamically creating pages in areas like the State and Territory-Specific Initiatives.

### Related ticket(s)

CMDCT-3858

---
### How to test
1. Run the Playwright tests `yarn test:e2e` or `yarn test:e2e-ui`
2. Run the "Creating a New Work Plan" test specifically. Notice that it is further serialized (as in, run one after the other) into "create" and "fill and submit."
3. Ideally, run the test more than once and ensure that it always passes. Trying to avoid flakiness!

### Notes

- I am not 100% committed to the POM as the way to write these tests. It's possible it won't serve us that well in the long run or is more complex that it needs to be for what few tests we are likely to write. But, for now, it seems to make the main test file readable.
- Serializing the work plan tests definitely slows it down a bit. And as we add more on top, we will want to keep trying to optimize as much as possible to keep things as speedy as we can under the circumstances.
- This journey has given me some ideas about exploring things like [storing authentication](https://playwright.dev/docs/auth) and possibly even intercepting the necessary [api keys for our requests](https://playwright.dev/docs/api/class-apirequestcontext) so we could quickly create/approve/archive reports like we did with Postman, but that is a journey for another day!

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/tests/e2e/utils/pageObjects/wp/wpInitiativeEvaluation.page.ts
+++ b/tests/e2e/utils/pageObjects/wp/wpInitiativeEvaluation.page.ts
@@ -1,4 +1,4 @@
-import { Locator, Page } from "@playwright/test";
+import { expect, Locator, Page } from "@playwright/test";
 import BasePage from "../base.page";
 
 export class WPEvaluationPlanPage extends BasePage {
@@ -44,6 +44,7 @@ export class WPEvaluationPlanPage extends BasePage {
     // Click the "No" radio option so we don't have to fill out conditional fields
     await radioGroup.getByLabel("No").click();
     await modal.getByRole("button", { name: "Save" }).click();
+    await expect(this.page.getByRole("alert")).not.toBeVisible();
     await modal.isHidden();
   }
 }
--- a/tests/e2e/utils/pageObjects/wp/wpInitiativeEvaluation.page.ts
+++ b/tests/e2e/utils/pageObjects/wp/wpInitiativeEvaluation.page.ts
@@ -21,9 +21,11 @@ export class WPEvaluationPlanPage extends BasePage {
         name: "Return to dashboard for this initiative",
       })
       .first();
-    this.addObjectiveButton = page.getByRole("button", {
-      name: "Add objective",
-    });
+    this.addObjectiveButton = page
+      .getByRole("button", {
+        name: "Add objective",
+      })
+      .first();
     this.modal = page.getByRole("dialog");
   }
 
--- a/tests/e2e/utils/pageObjects/wp/wpInitiativeEvaluation.page.ts
+++ b/tests/e2e/utils/pageObjects/wp/wpInitiativeEvaluation.page.ts
@@ -1,0 +1,47 @@
+import { Locator, Page } from "@playwright/test";
+import BasePage from "../base.page";
+
+export class WPEvaluationPlanPage extends BasePage {
+  public path = "/wp/state-or-territory-specific-initiatives/initiatives";
+
+  readonly page: Page;
+  readonly title: Locator;
+  readonly backButton: Locator;
+  readonly addObjectiveButton: Locator;
+  readonly modal: Locator;
+
+  constructor(page: Page) {
+    super(page);
+    this.page = page;
+    this.title = page.getByRole("heading", {
+      name: "State or Territory-Specific Initiatives: I. Define initiative",
+    });
+    this.backButton = page
+      .getByRole("button", {
+        name: "Return to dashboard for this initiative",
+      })
+      .first();
+    this.addObjectiveButton = page.getByRole("button", {
+      name: "Add objective",
+    });
+    this.modal = page.getByRole("dialog");
+  }
+
+  public async fillFields() {
+    const modal = this.page.getByRole("dialog");
+    const textAreas = await modal.getByRole("textbox").all();
+    const radioGroup = await modal.getByRole("group", {
+      name: "Does the performance measure include quantitative targets?",
+    });
+
+    // Fill all the text boxes in the modal
+    for (const textArea of textAreas) {
+      await textArea.fill("test");
+    }
+
+    // Click the "No" radio option so we don't have to fill out conditional fields
+    await radioGroup.getByLabel("No").click();
+    await modal.getByRole("button", { name: "Save" }).click();
+    await modal.isHidden();
+  }
+}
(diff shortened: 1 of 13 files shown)


tests/e2e/utils/pageObjects/wp/wpInitiativeEvaluation.page.ts
```diff
@@ -0,0 +1,50 @@
...
+      .getByRole("button", {
+        name: "Add objective",
+      })
+      .first();
+    this.modal = page.getByRole("dialog");
+  }
+
+  public async fillFields() {
+    const modal = this.page.getByRole("dialog");
+    const textAreas = await modal.getByRole("textbox").all();
```
jessabean, line 34: Using `.all()` opens us up to possibly flaky tests (see [note in documentation](https://playwright.dev/docs/api/class-locator#locator-all)), because it immediately returns the locators without necessarily waiting for the page to load. Paradoxically, it appears to be the only way to [iterate over locators](https://playwright.dev/docs/locators#do-something-with-each-element-in-the-list), so you will see that I've added some `expects` in these files to ensure that we're triggering certain actions before moving on, in an effort to tell the test runner to slow its roll.

tests/e2e/utils/pageObjects/wp/wpTransitionBenchmarkProjections.page.ts
```diff
@@ -0,0 +1,60 @@
...
+        await this.fillBenchmarks();
+        await drawer.getByRole("button", { name: "Save & Close" }).click();
+        await drawer.waitFor({ state: "hidden" });
+      }
+    }
+  }
+
+  public async fillBenchmarks() {
+    const drawer = this.page.getByRole("dialog");
+    await drawer.getByLabel("Yes").click();
```
gmrabian, line 54: is checking yes not duplicative of the above code in editPopulations?

tests/e2e/wp/create.spec.ts
```diff
@@ -1,52 +1,150 @@
...
+
+    // Transition Benchmark Strategy
+    await wpTransitionBenchmarksProjections.continueButton.click();
+    await wpTransitionBenchmarkStrategy.isReady();
+    await wpTransitionBenchmarkStrategy.fillTextFields();
+
+    // Initiatives Instructions
+    await wpTransitionBenchmarkStrategy.continueButton.click();
+    await wpInitiativesInstructions.isReady();
+    await wpInitiativesInstructions.fillFormFields();
```
gmrabian, line 95: looking at this flow I'm wondering if we'd consider renaming the "fill the page" option to something consistent across all pages?
`completeForm`, `fillForm`, `fillFormFields` etc.

gmrabian (approved):
This is great! I really like this pattern and it makes the test format pretty nice to layout. I think determining where to put some of the logic (iterating over initiatives, for example) is a bit flexible but I'm ok with the page objects being a bit more limited and having the test determine how and when to do things.

In regards to filling out fields I'm curious if there is any benefit to having reusable functions like
`fillText(textField, fillContent)`
`checkRadio(radioField, radioOption)` 
etc.
I do think playwright itself makes that all pretty clean though

tests/e2e/wp/create.spec.ts
jessabean: This is a very good point. I think I lost track of finesse as the days passed; I 100% support making the functions more consistent.

tests/e2e/utils/pageObjects/wp/wpInitiativeEvaluation.page.ts
benmartin-coforma: Interesting! Since this is iterating through fields in a modal, after the modal has loaded, I think it's reasonable for us to expect all those fields to have loaded.

I don't know if that's necessarily guaranteed by React's update loop or anything, but we're very much _not_ doing extra fetches to asynchronously add inputs after the modal is open. So it seems low-risk to me 👍

benmartin-coforma: I suppose, if we ever do run into flakiness here, the answer would be to explicitly list all the textboxes' expected labels? And then iterate over those, waiting to find (& fill) an input with each label. I've forgotten exactly what this modal looks like, so I can't remember how difficult that would be - if we'd be able to hardcode some labels, or if we'd need to auto-generate them based on the report period or what have you.

In either case, I'm not too worried about this.

tests/e2e/utils/pageObjects/wp/wpInitiativesInstructions.page.ts
```diff
@@ -0,0 +1,25 @@
...
+      name: "State or Territory-Specific Initiatives Instructions",
+    });
+  }
+
+  public async fillFormFields() {
+    const radioButtons = await this.page.getByLabel("No").all();
+
+    for (const radio of radioButtons) {
+      await radio.click();
+    }
```
benmartin-coforma, line 23: This is an instance where, without the page in front of me, it's hard for me to guess what is happening in the app. What questions are being answered "No" to? How many of them are there?

This isn't necessarily a bad thing. For one, almost everyone reading this test ever _will_ have the app in front of them, or at least be familiar enough to remember this page. And I only bring it up because it contrasts to so much of the rest of the code - like the easily-parseable `this.fundingSources.getByRole("button", { name: "Edit" }).click()`, or the supremely readable `fundingPage.addSourceButton.click()`

benmartin-coforma (approved):
I like it! The page models have definitely made the spec file easier to digest, and the code inside the page models is (weirdly, almost surprisingly) easy to read too. Within spitting distance of the "one line of code = one imperative English sentence" platonic ideal of scripting languages like JS.

I have some comments below, but they're just musings and not suggestions. You should merge this!

tests/e2e/utils/pageObjects/wp/wpTransitionBenchmarkProjections.page.ts
jessabean: Oh yeah you are right...I think this was a missed copy/move.

tests/e2e/utils/pageObjects/wp/wpInitiativesInstructions.page.ts
jessabean: I do think this could be named better in the long run, but this is the "special" initiatives page that has only 2 radio groups:

Screenshot 2024-09-30 at 3 35 57 PM

And in the interest of keeping these e2e tests as simple as possible, I've opted for the No option because I don't particularly care about the intricacies of making these two populations required for the initiative information in the following pages. Will make a note of renaming the variable to make this more clear!

PR #754, opened 2024-09-16
New API response pattern

benmartin-coforma:
### Description
This PR explores an alternative to Garrett's excellent idea in #752 , for streamlining our API handlers.

In short:
```ts
// Rather than constructing a response object on the fly (current state):
if (!allParamsValid) {
  return {
    status: StatusCodes.BadRequest,
    body: error.NO_KEY,
  };
}

// We will construct a response with a named helper:
if (!allParamsValid) {
  return badRequest(error.NO_KEY);
}
```

## How to Review
The best way to review this is probably commit-by-commit:
1. `Create a new type representing lambda results, with helpers`
   * Note the simplification of handler-lib, and the complexification of response-lib.
   * Note the change to `type LambdaFunction`: lambdas are now _required_ to return an `HttpResponse`
2. `Convert lambdas to use the new response helpers`
   * Pick your favorite handler and see how it has changed. My favorite is reports/fetch.ts
   * All the other changes are just that, times twelve. Safe to ignore if you're just trying to get the gist.
   * There were a few places where I changed the behavior slightly, such as banners/create.ts. I am pretty confident that none of these changes are consequential, beyond possibly streamlining a future API debugging session.
3. `Fix unit tests`
   * This commit was pure tedium. Safe to ignore.
4. `Rename Unauthorized -> Unauthenticated`
   * The only interesting changes are in response-lib.
   * This should probably have been in a separate PR, to be honest. I'm sorry, sometimes I can't help myself.
5. `Fix all the tests after that rename, whoops`
   * Ignore this one entirely please
   * Avert your eyes from my shame
6. `The rest of the commits`
   * OK maybe you should have gone file-by-file. Sorry! 😓 

### Related ticket(s)
n/a

---
### How to test
This PR touches almost every file in the API, and most changes have to do with requests that are invalid, for one reason or another. That, together with AWS' request signing, makes it very difficult to test from the UI. But, the expected behavioral changes are also very small - for example, changing a 403 response to a 409, for a situation that we don't expect to ever happen.

For that reason, I would say that the automated e2e tests are sufficient here - coupled, of course, with a thorough reading of the code changes.

### Notes
n/a

---
### Pre-review checklist

- [x] I have added [thorough](https://shorturl.at/aejkF) tests, if necessary
- ~[ ] I have updated relevant documentation, if necessary~
- [x] I have performed a self-review of my code
- [x] I have manually tested this PR in the deployed cloud environment

---
#### Review
[...]

PR #762, opened 2024-09-19
Improve error catching

gmrabian:
### Description

We had certain return blocks based on variables being undefined where they could not be undefined as a result of their flow. This fixes that logic with try/catches and builds on our [new response pattern](https://github.com/Enterprise-CMCS/macpro-mdct-mfp/pull/754)

### Related ticket(s)

CMDCT-3988

---
### How to test

All tests pass
Everything works as normal
If you resend a request locally with invalid fieldData or metadata if throws a 400 with the proper error message

### Notes

---
### Pre-review checklist


karla-vm (dismissed):

services/app-api/utils/auth/authorization.ts
```diff
@@ -51,17 +51,17 @@ export const isAuthenticated = async (event: APIGatewayProxyEvent) => {
...
+      isAuthenticated = true;
     } catch {
-      // verification failed - unauthenticated
-      isAuthenticated = false;
+      // verification failed - ignore and let it return false
     }
   }
 
-  return !!isAuthenticated;
+  return isAuthenticated;
```
benmartin-coforma, line 64: Suggestion, but non-blocking: My thought here is to go even further, if you've got the appetite. My priorities to improve maintainability are always to reduce (in order of importance): state, coupling, complexity, code. We could remove a tiny bit of state here by eliminating the `isAuthenticated` variable:

```ts
  try {
    await verifier.verify(event?.headers?.["x-api-key"]!);
    return true;
  } catch {
    return false;
  }
```

I am relying on `verifier.verify(undefined)` and `verifier.verify("")` to error out here, which I have not tested. Otherwise we may still need a guard clause up front:
```ts
  if (!event?.headers?.["x-api-key"]) {
    return false;
  }
```

benmartin-coforma (approved):
Sweet!

services/app-api/utils/auth/authorization.ts
gmrabian: Good stuff! I tested with `undefined` and with empty string. Both triggered the catch block.

gmrabian: I'll also note, for our own reassurance, this matches the [example the aws-jwt-verify package gives](https://www.npmjs.com/package/aws-jwt-verify), where any invalid token will throw an error. If it does not throw, it is considered to have passed validation.

karla-vm (approved):

benmartin-coforma (approved):

jessabean (approved):

PR #767, opened 2024-09-20
Fix disabled state for dashboard buttons

jessabean:
### Description
This is a hotfix for a bug I noticed: the disabled buttons and links on the work plan dashboard were extremely light and do not meet accessibility contrast standards

### Related ticket(s)

N/A

### How to test
1. Log in as a state user and create a work plan
2. Stay on the dashboard and see the disabled "Continue MFP Work Plan" and "Reset MFP Work Plan" buttons. They should be legible.

**Before**

Screenshot 2024-09-20 at 9 49 48 AM

**After**

Screenshot 2024-09-20 at 10 12 27 AM

### DIFF
--- a/services/ui-src/src/components/pages/Dashboard/DashboardPage.tsx
+++ b/services/ui-src/src/components/pages/Dashboard/DashboardPage.tsx
@@ -393,6 +393,7 @@ export const DashboardPage = ({ reportType }: Props) => {
           <Box sx={sx.callToActionContainer}>
             <Button
               type="submit"
+              variant="primary"
               disabled={isAddSubmissionDisabled()}
               onClick={() => openCreateReportModal()}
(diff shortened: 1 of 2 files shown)


bangbay-bluetiger (approved):
LGTM

jessabean:
> I'll give this an approval cause I actually disagree with WCAG here, but to my knowledge we don't actually need to meet color contrast for disabled elements. In WCAG it states that "Text or images of text that are part of an inactive user interface component, ... have no contrast requirement." So buttons that are disabled can be greyed out. Still, I don't think this hurts anything so take my 👍

oooh, TIL!! In any case, the extra light disabled state does not match the CMSDS disabled styles, so yes...I'll take your 👍🏻  lol

PR #771, opened 2024-09-25
Add text translate util

bangbay-bluetiger:
### Description

- Add a `translate` util that replaces any keys in `{{key}}` format within a string with defined values
- Use the util in `ExportedReportPage` with a LaunchDarkly flag
- Remove old LD flags in tests

### Related ticket(s)

CMDCT-3677

---
### How to test

1. Create a WP and SAR.
2. Verify the headings are as expected.
3. Unit tests also provide coverage.

### Notes

If we like this pattern, there are several other components where  we can use the util.

---
### Pre-review checklist

### DIFF
--- a/services/app-api/forms/sar.json
+++ b/services/app-api/forms/sar.json
@@ -182,6 +182,7 @@
             "intro": {
               "section": "Recruitment, Enrollment, and Transitions",
               "subsection": "Number of people who signed an MFP informed consent form in the reporting period",
+              "subsectionTitle": "Number of people who signed an MFP informed consent form in the {{reportingPeriod}}",
               "info": [
                 {
                   "type": "text",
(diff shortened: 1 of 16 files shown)


gmrabian (approved):
Changes look good to me. I'm curious to see the other use cases you have in mind. Is the advantage just of defining the desired string in a verbiage or json file rather than in the components?

bangbay-bluetiger:
> Changes look good to me. I'm curious to see the other use cases you have in mind. Is the advantage just of defining the desired string in a verbiage or json file rather than in the components?

Yes, defining the strings in verbiage files makes it easier to find them. Plus, instead of using string fragments where it isn't obvious they're being combined later, it's easier to see the complete string.

britt-mo (approved):
Looked good in the browser. I like this change, looks cleaner than using string interpolation in the component.

ntsummers1 (approved):

PR #779, opened 2024-10-03
Spike: Storing auth for playwright tests

jessabean:
### Description
Sets up a workflow to store and reuse authentication for specific users (admin, state user) that can be loaded and used in our tests. This means we can avoid running a login flow before each test that involves pages behind auth, and instead navigate directly to certain areas of the app with an already authenticated user.

### Related ticket(s)

CMDCT-3990

---
### How to test
1. From the root, run the Playwright tests with `yarn test:e2e` or `yarn test:e2e-ui`
2. Try to run all tests. They should all pass
3. Try to run individual tests. They should pass.
4. Before each/all of the tests, you should see the `utils/auth.setup.ts` file run first, as a dependency to all other tests.

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/tests/e2e/utils/auth.setup.ts
+++ b/tests/e2e/utils/auth.setup.ts
@@ -29,8 +29,8 @@ setup("authenticate as user", async ({ page }) => {
   const emailInput = page.getByRole("textbox", { name: "email" });
   const passwordInput = page.getByRole("textbox", { name: "password" });
   const loginButton = page.getByRole("button", { name: "Log In with Cognito" });
-  await emailInput.fill(statePassword);
-  await passwordInput.fill(stateUser);
+  await emailInput.fill(stateUser);
+  await passwordInput.fill(statePassword);
   await loginButton.click();
   await page.waitForURL("/");
   await page
--- a/tests/e2e/utils/auth.setup.ts
+++ b/tests/e2e/utils/auth.setup.ts
@@ -1,0 +1,43 @@
+import { test as setup } from "@playwright/test";
+
+import { adminPassword, adminUser, statePassword, stateUser } from "./consts";
+
+const adminFile = "playwright/.auth/admin.json";
+
+setup("authenticate as admin", async ({ page }) => {
+  await page.goto("/");
+  const emailInput = page.getByRole("textbox", { name: "email" });
+  const passwordInput = page.getByRole("textbox", { name: "password" });
+  const loginButton = page.getByRole("button", { name: "Log In with Cognito" });
+  await emailInput.fill(adminUser);
+  await passwordInput.fill(adminPassword);
+  await loginButton.click();
+  await page.waitForURL("/");
+  await page
+    .getByRole("heading", {
+      name: "View State/Territory Reports",
+    })
+    .isVisible();
+  await page.waitForTimeout(1000);
+  await page.context().storageState({ path: adminFile });
+});
+
+const userFile = "playwright/.auth/user.json";
+
+setup("authenticate as user", async ({ page }) => {
+  await page.goto("/");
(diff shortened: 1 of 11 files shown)


gmrabian (commented):
😮

jessabean:
## WUT is happening here?

### 1
We create a new folder where we will save browser local storage information during test runs, and we add that directory to our `.gitignore`

### 2
We create a `setup` function that is essentially another test. In this function we:

1. define the path where we want to store the auth information (cookies, origins)
2. go through the authentication steps as a user
3. tell Playwright to save the browser's cookies and local storage snapshot to that path
4. 
Screenshot 2024-10-07 at 3 36 40 PM


### 3
In our Playwright config, we set up a new project (which is basically a group of tests) and define it as anything that matches the filename `*.setup.ts`. Then we make that project a dependency of all of our other tests. This means that when running the e2e tests, the setup project will always run first, save login information for admin and state users, and then we have that information available to us for all the other tests.

Screenshot 2024-10-07 at 3 36 50 PM


### 4
We are already using fixtures and page objects in our tests. For some pages (e.g., `StateHomePage`) it is obvious that there is a specific user context for that page. So I went ahead and passed in the appropriate browser context when instantiating those pages. For other pages (e.g., any of the report pages), either user could see those pages and we'll want to set that context in the specific tests, so I left those alone.

Screenshot 2024-10-07 at 3 37 16 PM


Screenshot 2024-10-07 at 3 36 59 PM

### 5
Now in our tests, we can define multiple contexts for admin and regular users, and load up the saved browser info for each (these files have already been written as part of the setup function all the way back in step 2). These contexts will be ready to go when we need to instantiate new report page objects during the tests. In the case of `AdminHomePage` and `StateHomePage` we already set those contexts in the fixtures. This difference can be seen in lines 31 and 33 in the second screenshot, below.

Screenshot 2024-10-07 at 3 37 36 PM


 
Screenshot 2024-10-07 at 3 38 13 PM

gmrabian (approved):
Cool stuff! It's nice to see all these features coming together.

some non-blocking comments:

tests/e2e/utils/auth.setup.ts
```diff
@@ -0,0 +1,43 @@
...
+  await emailInput.fill(adminUser);
+  await passwordInput.fill(adminPassword);
+  await loginButton.click();
+  await page.waitForURL("/");
+  await page
+    .getByRole("heading", {
+      name: "View State/Territory Reports",
+    })
+    .isVisible();
+  await page.waitForTimeout(1000);
```
JonHolman, line 21: I would prefer to find something to wait for that would signal what we're waiting for is done, as set times are unreliable.

tests/e2e/wp/create.spec.ts
```diff
@@ -1,18 +1,60 @@
...
+});
 
 test.describe("Creating a new Work Plan", () => {
-  test("State user can create a Work Plan", async ({
-    page,
-    stateHomePage,
-    wpDashboard,
-  }) => {
+  test("State user can create a Work Plan", async ({ page }) => {
+    // TODO: migrate these functions to use the save auth model
```
JonHolman, line 51: Is there another ticket for this TODO?

tests/e2e/utils/auth.setup.ts
jessabean: I agree on principle! In the above 2 lines we are waiting for elements to be visible (and they would only be visible if logged in), but in this case it appears to not be enough...it's like the browser just needs a beat to completely finish saving the cookies before moving to the next test.

tests/e2e/wp/create.spec.ts
jessabean: Will add one!

PR #787, opened 2024-10-15
Return component JSX instead of calling function

jessabean:
### Description
Fixes a bug where resizing the viewport to a mobile size on the Review & Submit page produces an error and the page does not render.

Took the opportunity to do a little refactoring so I could write a unit test for the viewport resizing.

https://github.com/user-attachments/assets/d53c2466-7d57-4232-a153-9a697940c885

---
### How to test
1. Start a work plan
2. Go to Review & Submit page
3. Resize the window to be very smol, and back to big again.
4. You should not see any console errors, and the page should render as expected


gmrabian (approved):
hilarious bug. great fix.

PR #791, opened 2024-10-18
Upgrade amplify to v6

gmrabian:
### Description

This change upgrades Amplify from v5 to v6 

The upgrade work was previously done in [CARTS](https://github.com/Enterprise-CMCS/macpro-mdct-carts/pull/139616) but was reverted due to an [issue](https://github.com/aws-amplify/amplify-js/issues/13310) that has since been resolved by the maintainers.

Work already [done in MCR](https://github.com/Enterprise-CMCS/macpro-mdct-mcr/pull/11876)

For more info on the oauth amplify listener in root, see this [PR](https://github.com/Enterprise-CMCS/macpro-mdct-mcr/pull/11897)

References:
[upgrade guide](https://docs.amplify.aws/gen1/react/build-a-backend/troubleshooting/migrate-from-javascript-v5-to-v6/)

### Related ticket(s)

CMDCT-4052

---
### How to test

Everything in the app works as expected. A short list of everything:
- all api calls
- auth
- timeout
- template file download

---
### Pre-review checklist


services/ui-src/src/utils/other/print.ts
```diff
@@ -21,15 +21,18 @@ export const printPdf = async () => {
...
-  const request = {
+  const options = {
     headers: { ...requestHeaders },
     body: { encodedHtml: base64String },
   };
+  const path = `/print_pdf`;
   let response;
   if (config.DEV_API_URL) {
-    response = await API.post("mfpDev", `/print_pdf`, request);
+    const { body } = await post({ apiName: "mfpDev", path, options }).response;
```
bangbay-bluetiger, line 31: Combine these?
```
const apiName = config.DEV_API_URL ? "mfpDev" : "mfp";
const { body } = await post({ apiName, path, options }).response;
const response = (await body.json()) as string; // remove `let response`
```

bangbay-bluetiger (commented):
Looks good. Minor suggestion.

tests/cypress/e2e/accessibility/profile.cy.js
```diff
@@ -12,6 +13,7 @@ describe("Profile Page - Accessibility Test", () => {
     cy.authenticate("stateUser");
 
     cy.visit("/profile");
+    cy.wait(1000);
```
JonHolman, line 16: Can we wait for something instead of waiting for 1000 ms?

gmrabian: I'd love to 😭 cypress doesn't have many options for that though. I tried awaiting an api call that should happen on that page but it still gets the error. not sure what is causing it.

JonHolman: Ah.  I haven't done much with cypress.  I've never tried this, but might be worth taking a look https://www.npmjs.com/package/cypress-wait-until

bangbay-bluetiger (dismissed):

services/ui-src/src/components/reports/ReportProvider.tsx
```diff
@@ -93,6 +93,7 @@ export const ReportProvider = ({ children }: Props) => {
       return result;
     } catch {
       setError(reportErrors.GET_REPORT_FAILED);
+      return;
```
JonHolman, line 96: extreme nit, we don't need this return here.  Right?

JonHolman (dismissed):

services/ui-src/src/components/reports/ReportProvider.tsx
britt-mo: @JonHolman without that `return` you'll see a typescript error `Not all code paths return a value.`

JonHolman: Thank you, @britt-mo

tests/cypress/e2e/accessibility/profile.cy.js
jessabean: @gmrabian you might be able to replace this with a `cy.get("h1")` (could replace with another selector like "main"). I think this will ensure that the page has finished rendering before making the a11y assertions.

gmrabian:
you can still review but we need to figure out the IDM integration before merging

jessabean (approved):

tests/cypress/e2e/accessibility/profile.cy.js
gmrabian: Changed and it worked!

karla-vm (approved):

karla-vm (approved):

jessabean (approved):

PR #792, opened 2024-10-18
More Playwright tinkerings

jessabean:
### Description

- Use the stored auth context to archive all plans
- Move the archive plans function to the dashboard page object
- Small tinkering with awaits and expects to try to work with this checkbox situation ¯\\(°_o)/¯

### Related ticket(s)

CMDCT-

---
### How to test

- Run the playwright tests
- Hopefully...they pass?


gmrabian (approved):

gmrabian (approved):
tests pass! I'm wondering if we can use the blur function that doesn't use a timeout from cmsds?
it looks like timeout is only for the `onComponentBlur` function. Any way we could just use `onBlur`? [cmsds component](https://github.com/CMSgov/design-system/blob/main/packages/design-system/src/components/ChoiceList/ChoiceList.tsx#L97-L111)

PR #798, opened 2024-10-31
Improve error handling in seed script

bangbay-bluetiger:
### Description

The seed script was erroring silently if it couldn't create a WP or SAR and would terminate the script. These updates print errors to console and leaves the script running.

Also:
- Split changing reporting period to separate year and period options
- Remember selected reporting period when script goes back to main menu
- Add "Back to menu" options when in WP and SAR submenus

---
### How to test

1. Start MFP locally
2. Run `yarn db:seed`
3. Confirm WPs and SARs can be created for reporting period
4. Confirm if WPs and SARs already exist for reporting period, receive an accurate error message

---
### Pre-review checklist

### DIFF
--- a/services/app-api/db/helpers.ts
+++ b/services/app-api/db/helpers.ts
@@ -14,3 +14,6 @@ export const createdLog = (obj: any, action: string, type: string): void => {
 export const expandedLog = (json: any): void => {
   console.log(JSON.stringify(json, null, 2));
 };
+
+export const generateReportingPeriod = (year: number, period: number) =>
+  `${year} Period ${period}`;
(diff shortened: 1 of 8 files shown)


jessabean (dismissed):
👍🏻

gmrabian (commented):
I changed reporting year, created a WP and SAR, then attempted to change reporting year again and it didn't change

Screenshot 2024-11-01 at 12 05 54 PM

I also noticed that if you run GetBanner when there is no banner it exits (probably on an error?)

Love the updates! These are sort of edge cases and not necessary for the expected changes in this PR

gmrabian (approved):
looks good! played around locally and the error catches/messaging worked nicely 🚀

PR #801, opened 2024-11-06
Refactor amplify methods to single file

bangbay-bluetiger:
### Description

[Amplify](https://github.com/aws-amplify/amplify-js) typically has breaking changes with each version upgrade. To mitigate how much refactoring we have to do with each upgrade, we should use generic helpers and call amplify methods from a single file, as much as possible.

Other changes in this PR:

- Add a 204 response code for banner not found so it doesn't give an error console output
- Convert deprecated `atob` and `btoa`

### Related ticket(s)

CMDCT-4037

---
### How to test

- Everything in the app works as expected. A short list of everything:
  - all api calls
  - auth
  - timeout
  - template file download
- Automated tests pass

### Notes

Left in uses of `AnyObject` where there's a `TODO` to replace them. Because of that, I needed to cast to `ReportShape` in a couple of places. I put `TODO`s as a reminder to remove the casting when the `AnyObject` conversion happens.

---
### Pre-review checklist


karla-vm (approved):

services/app-api/handlers/banners/fetch.ts
```diff
@@ -2,13 +2,17 @@ import handler from "../handler-lib";
...
     return badRequest(error.NO_KEY);
   }
-  const bannerId = event?.pathParameters?.bannerId!;
   const banner = await getBanner(bannerId);
-  return ok(banner);
+  if (banner) {
+    return ok(banner);
+  }
+  // Instead of 404, return 204 for silent error
+  return noContent();
```
gmrabian, line 17: discussion: i believe the reason we originally used a 404 is because the banner fetch is fetching on a specific id. granted, it's the id we use for all banners. i think we've flipped between 204 and 404 before. i am fine with 204, just calling it out in case anyone wants to get api nerdy with me 🤓

gmrabian (approved):
great work! ⭐

services/app-api/handlers/banners/fetch.ts
bangbay-bluetiger: Right now, the banner doesn't 404. It gives a 200 even if it doesn't exist (uses undefined body). So, in my comment, I meant that a 404 is the more logical choice here, but it seems 204 is more in line with what we were doing with the 200.

PR #826, opened 2024-12-12
Remove unnecessary prop drilling

gmrabian:
### Description

Stumbled upon some prop drilling which I believe had no benefit. This removes it, simplifies context access, and adds a test!

---
### How to test

Tests pass
Logout still works as a user


britt-mo (approved):
Tested logout locally

PR #827, opened 2024-12-12
CMDCT-3860: Playwright test for resubmit WP

jessabean:
### Description
Adds test for resubmitting a work plan

### Related ticket(s)

CMDCT-3860

---
### How to test
- run the playwright tests
- they should pass

### DIFF
--- a/tests/e2e/wp/create.spec.ts
+++ b/tests/e2e/wp/create.spec.ts
@@ -210,4 +210,18 @@ test.describe("Creating a new Work Plan", () => {
     await modal.getByRole("button", { name: "Return to dashboard" }).click();
     await expect(unlockButton).toBeDisabled();
   });
+
+  test("State user can resubmit a work plan", async () => {
+    await wpDashboard.goto();
+    await wpDashboard.firstReport.getByRole("button", { name: "Edit" }).click();
+    await wpGeneralInformation.isReady();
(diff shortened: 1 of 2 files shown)


tests/e2e/wp/create.spec.ts
```diff
@@ -210,4 +210,18 @@ test.describe("Creating a new Work Plan", () => {
     await modal.getByRole("button", { name: "Return to dashboard" }).click();
     await expect(unlockButton).toBeDisabled();
   });
+
+  test("State user can resubmit a work plan", async () => {
+    await wpDashboard.goto();
+    await wpDashboard.firstReport.getByRole("button", { name: "Edit" }).click();
```
britt-mo, line 216: Do we care about writing out the selector in this file vs in `wpDashboard.page.ts`?

britt-mo (approved):
Test  passed, just a nit comment

bangbay-bluetiger (approved):

gmrabian (approved):

tests/e2e/wp/create.spec.ts
jessabean: In general I would lean toward only including selectors in the page object if we know they are going to be used more than once. I see that we are actually using the Edit button here twice, so it could be a candidate to move. I'll follow up on that!

PR #874, opened 2025-02-10
CMDCT-4325 - STAGE_PREFIX is not real

peoplespete:
### Description
STAGE_PREFIX was never being used. So let's remove it.

### Related ticket(s)
CMDCT-4325

---
### How to test
Note that build didn't break after it was removed from the code.

### Important updates
NA

---
### Author checklist

---


jessabean (approved):
👍🏻 Built and runs on local env and the cloudfront URL

PR #888, opened 2025-03-05
Playwright updates

JonHolman:
### Description
Playwright updates - Several minor style updates - refactored authentication

### Related ticket(s)

CMDCT-

---
### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---


tests/playwright/utils/auth.setup.ts
```diff
@@ -1,43 +1,109 @@
...
-    .getByRole("heading", {
-      name: "Money Follows the Person (MFP) Portal",
-    })
-    .isVisible();
-  await page.waitForTimeout(1000);
-  await page.context().storageState({ path: userFile });
-});
+  await waitForJwtResponse(page);
+  await expect(page.getByRole("heading", { name: finalHeading })).toBeVisible();
+}
```
jessabean, line 109: I think we can keep the consts for the file path, but I don't think we need the rest of the changes in this file. This file should be pretty bare bones, and the `setup` keyword is intentional and necessary. This is the signal that tells the test runner to run this file before all the other spec files, along with the config. You can see that we are following the [Playwright docs on storing authentication state](https://playwright.dev/docs/auth#multiple-signed-in-roles).

tests/playwright/utils/consts.ts
```diff
@@ -3,6 +3,11 @@ export const adminPassword = process.env.SEED_ADMIN_USER_PASSWORD!; // pragma: a
 export const stateUser = process.env.SEED_STATE_USER_EMAIL!;
 export const statePassword = process.env.SEED_STATE_USER_PASSWORD!; // pragma: allowlist secret
 
+/** The path to the file containing admin browser session data */
+export const adminAuthPath = "playwright/.auth/admin.json";
+/** The path to the file containing state user browser session data */
+export const stateUserAuthPath = "playwright/.auth/user.json";
+
```
jessabean, line 10: 👍🏻

tests/playwright/utils/logging.ts
```diff
@@ -0,0 +1,66 @@
+/* eslint-disable no-console */
```
jessabean, line 1: As discussed in the working session, not sure we need this? Is it purely for debugging?

jessabean (commented):
TL;DR:

- I think 👍🏻 on pulling over file path consts
- Yay for unique test names
- I _don't_ think we should or need to change the auth setup
- Not sure we need the logging stuff, unless you've got a strong argument for if it would be helpful in a CI environment?

tests/playwright/utils/logging.ts
JonHolman: Yes, purely for information while investigating, I'll remove.

tests/playwright/utils/auth.setup.ts
JonHolman: Thanks.  I put the test as setup back.  The big change here was checking the context if it was already logged in prior to logging in, as that speeds things up.

jessabean:
Closing and pulling some of the changes into #889

PR #892, opened 2025-03-14
Update yarn locks

bangbay-bluetiger:
### Description

This PR cleans up the node packages installed in MFP.

- Update `browserslist` and its dependencies
  - `caniuse-lite`
  - `electron-to-chromium`
  - `node-releases`
  - `update-browserslist-db`
    - `escalade`
    - `picocolors`
- Combine minor versions of packages to resolve to the latest minor version already installed in MFP
  - This should result only in removal of old versions and dependencies, and no installation of new packages
    - One way to confirm this is in the diff, no `version` lines are changed except for the packages related to `browserslist`
  - In `ui-src` and `app-api`, cleanup particularly involves many `@babel` packages
  - In `database`, cleanup is of a redundant `@aws-sdk/client-dynamodb` package
- Resolve peer dependency warnings related to  `ts-node`, `@testing-library/user-event`, `@axe-core/playwright` by adding them to `package.json`
  - Leave intentional warnings related to package versions we're using

This results in a overall decrease in size of the `node_modules` folders from `main` -> this branch:

**Locally:**

- root: 618MB -> 517MB
- `app-api`: 330MB -> 303MB
- `database`: 137MB -> 33MB
- `topics`: 70MB -> 70MB, no change
- `ui-auth`: 19MB -> 17MB
- `ui-src`: 988MB -> 835MB
- total: 2162MB -> 1775MB, a 17.9% decrease

**Production:**

- root: 64MB -> 64MB, no change
- `app-api`: 65MB -> 55MB
- `database`: 50MB -> 24MB
- `topics`: 1.9MB -> 1.9MB, no change
- `ui-auth`: 19MB -> 17MB
- `ui-src`: 551MB -> 400MB
- total: 750.9MB -> 561.9MB, a 25.17% decrease

---
### How to test

1. Run MFP locally
2. Everything should work as normal
3. Run `yarn test` in `app-api` and `ui-src`

### Notes

- Installation sizes were calculated with [this script](https://gist.github.com/bangbay-bluetiger/3c031d9578d32d4b6b754003e967db1a)
- If you get an error for missing `patch-package` from doing a production install, you can install `patch-package` globally and then add it to your path
  - `yarn global add patch-package`
  - Add `~/node_modules/.bin` to your `PATH` in your shell `rc` file
---
### Pre-review checklist

- [ ] I have added [thorough](https://shorturl.at/aejkF) tests, if necessary
- [ ] I have updated relevant documentation, if necessary
- [ ] I have performed a self-review of my code
[...]

PR #894, opened 2025-03-18
Bump AWS SDK version numbers to 3.758.0

ntsummers1:
### Description

See title

---
### How to test

Run through the app, double check it works!

### Notes

All aws-sdk version to 3.758!

---
### Pre-review checklist

---

### DIFF
--- a/services/database/yarn.lock
+++ b/services/database/yarn.lock
@@ -120,115 +120,115 @@
     tslib "^2.6.2"
     uuid "^9.0.1"
 
-"@aws-sdk/client-dynamodb@^3.731.1":
-  version "3.731.1"
-  resolved "https://registry.yarnpkg.com/@aws-sdk/client-dynamodb/-/client-dynamodb-3.731.1.tgz#d61f99e5fd33ceb5e5bd7fa6bb3de11c17bc3827"
-  integrity sha512-rdvGbusvbcp4NB1gAr9OGYD9cbyxXNAAaq3BopFMmtqUW1vhD+8DXcQYnZBAu93qjr8Nep1qDtysoIp+AeSOjg==
+"@aws-sdk/client-dynamodb@^3.758.0":
+  version "3.767.0"
+  resolved "https://registry.yarnpkg.com/@aws-sdk/client-dynamodb/-/client-dynamodb-3.767.0.tgz#93674371e48c32706838750b96fba71d3acd33cf"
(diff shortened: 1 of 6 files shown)


gmrabian (approved):
Didn't see any issues in deployed env!

Some conflicts with main. May want to double check this doesn't conflict/undo any of [Bangbay's yarn.lock updates](https://github.com/Enterprise-CMCS/macpro-mdct-mfp/pull/892)

jessabean (dismissed):
🙌🏻  Thank you

services/database/yarn.lock
```diff
@@ -119,229 +119,278 @@
     tslib "^2.6.2"
     uuid "^9.0.1"
 
-"@aws-sdk/client-s3@^3.731.1":
-  version "3.732.0"
-  resolved "https://registry.yarnpkg.com/@aws-sdk/client-s3/-/client-s3-3.732.0.tgz#936e8f0ce66cde332195cc77542c0ea01cae25a9"
-  integrity sha512-ITPcG40qdiLXZRvNQ8V3u4yB16afcdIabdRBN6Blba31rk0MfeBGWwah0+lLSTFo1ZIrIQvBl6PAQ7mO0mkKLg==
+"@aws-sdk/client-dynamodb@^3.758.0":
```
bangbay-bluetiger, line 122: `@aws-sdk/client-dynamodb@^3.428.0` and `@aws-sdk/client-dynamodb@^3.758.0` can be combined to resolve to the same version `3.767.0`. Sometimes yarn can do it automatically, but you can combine the lines manually. Also have been trying out [yarn-deduplicate](https://www.npmjs.com/package/yarn-deduplicate) to do the deduping.

ntsummers1: This is a great call out. We've been trying to keep all the aws-sdk versions on the same one and only recently has amazon started to deviate. For instance, the newest version of client-s3 is 3.758 which is why its the version chosen here. I admit my deduplication knowledge is little to non-exsistant, if I wanted to keep the version as 3.758 do you think I could reduce it?

bangbay-bluetiger: If we want the version to be exactly `3.758.0`, we could either set the version in `package.json` to "@aws-sdk/client-dynamodb": "3.758.0" or add it as a resolution. Otherwise, the latest compatible version to `^3.758.0` is `3.767.0`, so that's why yarn auto picked it.

`serverless-dynamodb@^0.2.5` is the package that wants `@aws-sdk/client-dynamodb@^3.428.0`. If we set `3.758.0` as our version, yarn won't be smart enough to just combine the two and will install both `3.758.0` and `3.767.0` (as the latest compatible with `^3.428.0`)  unless we use resolutions.

ntsummers1: Gotcha. Should hopefully be de-dupped now!

PR #897, opened 2025-03-24
Allow admin users to create and manage multiple banners

gmrabian:
### Description

- Allow multiple banners to exist at once in database
- Allow admins to view all existing banners, active, expired, and scheduled
- Add logic to filter and sort banners so that applicable date range with most recent creation date is active
- Add logic to show an error and not create a banner when the new banner dates overlap with existing banners
- Adjust tests and api logic to fit pattern

Most of the new logic is in `AdminBannerProvider` and `AdminPage`

Pattern already in use in [MCR](https://github.com/Enterprise-CMCS/macpro-mdct-mcr/pull/12057)

### Related ticket(s)

CMDCT-4442

---
### How to test

- Log in as an admin
- Go to `/admin`
- Create a banner
- Create another banner
- Notice they both display in a list at the top
- Experiment with creating and deleting banners
  - Note that if you try to create a banner where either the start date or the end date fall within the existing banner date ranges, it will throw an error on submit and not create the banner
- Visit the homepage to see what banner is active
  - It should always be the banner where today's date falls within the range and with the most recent creation date

### Notes
Some of the yarn.lock files didn't match `main`. Since all those changes remove dependencies, and we recently had an effort to limit duplicates, I thought it was safe to include here.

---
### Pre-review checklist

---
#### Review
---


britt-mo (approved):
Really thorough work, great job @gmrabian 

I tested locally and everything worked as expected (banners with overlapping dates would trigger an error, only the current date's banner would display on the report, and delete worked correctly)

bangbay-bluetiger (approved):

PR #903, opened 2025-03-27
CMDCT-4362 - cdk transition sls side DO NOT MERGE

JonHolman:
### Description
This is the changes we need to make to the serverless setup before we convert to cdk.

### Related ticket(s)
CMDCT-4362

---
### How to test
You can double check that we haven't left out any resources that could or should be retained. At a minimum we need Cognito Userpool, most s3 buckets, all Dynamo Tables, and Cloudfront Distribution.
The tests don't pass because this change breaks the app. That's fine when we are in the moment of transition but that's one reason we should never merge this :)

### Notes

---
### Pre-review checklist

- [X] I have added [thorough](https://shorturl.at/aejkF) tests, if necessary
- [X] I have updated relevant documentation, if necessary
- [X] I have manually tested this PR in the deployed cloud environment

---
#### Review

#### Security

---


services/app-api/serverless.yml
```diff
@@ -31,9 +31,7 @@ custom:
       - "SizeRestrictions_BODY"
   serverlessTerminationProtection:
     stages:
-      - main
-      - val
-      - production
+      - nonsense
```
JonHolman, line 34: had to put something there to disable termination protection on all other stages.

services/ui-src/tsconfig.json
```diff
@@ -16,7 +16,7 @@
     "module": "esnext",
     "baseUrl": "src",
     "esModuleInterop": true,
-    "forceConsistentCasingInFileNames": true,
+    "forceConsistentCasingInFileNames": false,
```
JonHolman, line 19: Hopefully we don't actually need this change, maybe it was other changes that were merged while I was testing, but this workaround allowed me to resume my testing.

src/run.ts
JonHolman, line 1: To prepare to run destroy on our real persistent stages, for the CDK migration, commented out delete_topics and added safety checks to ensure stateful and auditing resources and their contents are being retained prior to stack destruction.

services/ui-src/tsconfig.json
peoplespete: Yeah I don't really know that we need this but not a huge deal for me.

services/ui/serverless.yml
```diff
@@ -190,9 +189,9 @@ resources:
         LoggingConfiguration:
           DestinationBucketName: ${env:LOGGING_BUCKET, ssm:/configuration/${self:custom.stage}/s3/accessLogsBucket, ssm:/configuration/default/s3/accessLogsBucket}
           LogFilePrefix: ${env:LOGGING_BUCKET, ssm:/configuration/${self:custom.stage}/s3/accessLogsPrefix, ssm:/configuration/default/s3/accessLogsPrefix}
-      DeletionPolicy: Delete
     LoggingBucketPolicy:
       Type: "AWS::S3::BucketPolicy"
+      DeletionPolicy: Retain
```
peoplespete, line 194: Does this contain something stateful? I was hoping we didn't have to retain policy-esque stuff.

yarn.lock
peoplespete, line 1: Didn't expect this to change without accompanying package.json changing.

services/ui/serverless.yml
JonHolman: I don't think we can delete historical logs simply because we migrated.  I'm sure we have a retention policy of logs.

peoplespete (dismissed):
DON"T MERGE bc it is the sls branch :)
But other than a couple tiny things, it looks great!

services/ui/serverless.yml
JonHolman: oh!  This.  After I deployed, CDK didn't use this bucket... so we need to keep this policy so that this historical log bucket does get a security finding for allowing non-SSL traffic.

yarn.lock
JonHolman: it shouldn't.... but it did... :(

src/run.ts
```diff
@@ -150,13 +248,78 @@ async function destroy_stage(options: {
...
+    bucketsToSkip: [
+      `database-${options.stage}-sar`,
+      `database-${options.stage}-wp`,
+      `ui-${options.stage}-cloudfront-logs-${accountId}`,
+      `${accountId}-ui-${options.stage}-waflogs`,
+    ],
   });
 
-  await delete_topics(options);
+  // await delete_topics(options);
```
gmrabian, line 316: are we skipping this or just enabling later?

gmrabian (commented):
Things here look good to me. What's the next step?

JonHolman:
> Things here look good to me. What's the next step?

Thanks @gmrabian this branch will be deployed to a stage to prepare it to migrate to CDK.  The import instructions document the process.

PR #917, opened 2025-04-24
Generate a .env.seed with API_URL

bangbay-bluetiger:
### Description

With the CDK conversion, `API_URL` is now dynamically generated, so the key was removed from `.env`. The `yarn db:seed` script was using that key, so we need a another way to pass that value to the script. This PR creates a `.env.seed` file when you run MFP locally.

---
### How to test

1. Run MFP locally
2. Open another terminal window and run `yarn db:seed`
3. All seed commands should work

### Notes

---
### Pre-review checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ build_dev
 tsconfig.tsbuildinfo
 services/uploads/local_buckets
 /.env
+/.env.seed
 tests_output
 *.log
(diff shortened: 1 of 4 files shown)


JonHolman (approved):

britt-mo (approved):
Tested, seed commands worked as expected

PR #924, opened 2025-05-01
CMDCT-4622 gets cloudfront logging to stop breaking destroys for ephemeral apps

peoplespete:
### Description
This change will mean that we no longer have issues destroying branches which destroys the stack.

### Related ticket(s)
CMDCT-4622

---
### How to test
The app still works the same.

### Notes
NA

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/deployment/stacks/data.ts
+++ b/deployment/stacks/data.ts
@@ -1,5 +1,10 @@
 import { Construct } from "constructs";
-import { aws_dynamodb as dynamodb, aws_s3 as s3, Aws } from "aws-cdk-lib";
+import {
+  aws_dynamodb as dynamodb,
+  aws_s3 as s3,
+  Aws,
+  RemovalPolicy,
+} from "aws-cdk-lib";
 import { DynamoDBTable } from "../constructs/dynamodb-table";
 
 interface CreateDataComponentsProps {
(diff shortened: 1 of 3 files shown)


deployment/stacks/data.ts
```diff
@@ -59,21 +64,25 @@ export function createDataComponents(props: CreateDataComponentsProps) {
...
     bucketName: `database-${stage}-sar`,
     encryption: s3.BucketEncryption.S3_MANAGED,
-    versioned: true,
     blockPublicAccess: s3.BlockPublicAccess.BLOCK_ALL,
     serverAccessLogsBucket: loggingBucket,
     serverAccessLogsPrefix: `AWSLogs/${Aws.ACCOUNT_ID}/s3/`,
+    versioned: true,
     enforceSSL: true,
+    removalPolicy: isDev ? RemovalPolicy.DESTROY : RemovalPolicy.RETAIN,
+    autoDeleteObjects: isDev,
```
peoplespete, line 73: @JonHolman Does this seem like a reasonable change? I just noticed an issue when I deployed the same branch twice the buckets were still there so it had a name conflict.

deployment/stacks/ui-auth.ts
```diff
@@ -142,7 +142,8 @@ export function createUiAuthComponents(props: CreateUiAuthComponentsProps) {
   const userPoolDomain = new cognito.UserPoolDomain(scope, "UserPoolDomain", {
     userPool,
     cognitoDomain: {
-      domainPrefix: userPoolDomainPrefix ?? `${stage}-login-user-pool-client`,
+      domainPrefix:
+        userPoolDomainPrefix ?? `${project}-${stage}-login-user-pool-client`,
```
peoplespete, line 146: This is the thing to prevent global name collisions since the url is global.

deployment/stacks/ui.ts
```diff
@@ -55,26 +55,38 @@ export function createUiComponents(props: CreateUiComponentsProps) {
...
+  if (!isDev) {
+    // this bucket is not created for ephemeral environments because the delete of the bucket often fails because it doesn't decouple from the distribution gracefully
+    // should you need to test these parts of the infrastructure out the easiest method is to add your branch's name to the isDev definition in deployment-config.ts
+    const logBucket = new s3.Bucket(scope, "CloudfrontLogBucket", {
+      bucketName: `ui-${stage}-cloudfront-logs-${Aws.ACCOUNT_ID}`,
+      encryption: s3.BucketEncryption.S3_MANAGED,
+      publicReadAccess: false,
+      blockPublicAccess: s3.BlockPublicAccess.BLOCK_ALL,
+      objectOwnership: s3.ObjectOwnership.BUCKET_OWNER_PREFERRED,
+      removalPolicy: RemovalPolicy.RETAIN,
```
JonHolman, line 70: Due to your comment, I above this bucket declaration, I feel this line should remain:
```suggestion
      removalPolicy: isDev ? RemovalPolicy.DESTROY : RemovalPolicy.RETAIN,
      autoDeleteObjects: isDev,
```

JonHolman (commented):
I like it, just a small suggestion and tagged @bangbay-bluetiger and @jessabean to confirm we can empty dev ephemeral branches WP and SAR buckets.

deployment/stacks/ui.ts
peoplespete: I just didn't like having anything inside an if block that is unreachable. Since this is inside a !isDev if statement, Removal.DESTROY is not reachable. Also it means that autoDeleteObjects will always be false which is the default.

jessabean (approved):

deployment/stacks/data.ts
peoplespete: Got confirmation from @jessabean that it is ok to delete the buckets as long as we don't for main/val/prod stacks.

jessabean: Yeah I think that's fine!

PR #942, opened 2025-05-29
Update optional field schemas

gmrabian:
### Description

Update frontend validation schemas for optional fields to:
- accept `null` values
- convert empty strings to `null` values

Fields where no value is entered or a previous value was removed now save as `null` instead of empty string. Is this something we need to communicate to DataConnect?

I did not update the schemas in the backend because they only receive a singular value. They do not have to determine validation to changing values. 

### Related ticket(s)

CMDCT-4444

---
### How to test

[Deployed env](https://d18uja5w5t1kux.cloudfront.net/)

Admin user
- Log in as an admin and navigate to `/admin`
- Enter text in the banner form's Link field and then delete it
  - Verify validation goes away when the field is empty
  - Verify you can still submit the form when link had text entered and then removed

State user
- Run MFP locally
- Run `yarn db:seed` and create an approved WP in Period 2
- Create a SAR and navigate to the HCBS RE&T section
- Enter numbers for some of the fields
  - Verify you can remove the numbers and still save
  - Verify the values are saved as `null` when removed
  - Verify you can enter and then delete text and validation goes away

---
### Pre-review checklist

---
#### Review
---

### DIFF
--- a/services/ui-src/src/utils/validation/schemas.test.ts
+++ b/services/ui-src/src/utils/validation/schemas.test.ts
@@ -1,5 +1,15 @@
 import { MixedSchema } from "yup/lib/mixed";
-import { number, ratio, validInteger } from "./schemas";
+import {
+  dateOptional,
+  emailOptional,
+  number,
+  numberOptional,
+  ratio,
+  textOptional,
+  urlOptional,
+  validInteger,
+  validIntegerOptional,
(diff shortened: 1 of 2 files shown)


bangbay-bluetiger (dismissed):
Works as expected. Non-blocking code clean up suggestions.

PR #977, opened 2025-07-02
Fix focus indicator

gmrabian:
### Description

Focus indicator needs to be switched to our new mdct styling prefix

How to test:
tab through app and verify pink-ish focus indicator shows!


cy-by (approved):
👯 


https://github.com/user-attachments/assets/55034211-2e15-4629-b93c-c33479122192

PR #1053, opened 2025-09-15
Clean up CDK code by using grant helper functions where possible instead of creating IAM policies

JonHolman:
### Description

Clean up CDK code by using grant helper functions where possible instead of creating IAM policies

### Related ticket(s)

CMDCT-4980

---

### Notes

---

### Pre-review checklist

---

#### Review

#### Security

---


deployment/stacks/api.ts
```diff
@@ -107,142 +130,99 @@ export function createApiComponents(props: CreateApiComponentsProps) {
...
     stackName: `${service}-${stage}`,
     api,
     environment,
-    additionalPolicies,
+    tables,
     isDev,
   };
 
-  new Lambda(scope, "createBanner", {
+  const createBanner = new Lambda(scope, "createBanner", {
```
peoplespete, line 145: Any chance we could set this up like seds so that we don't have to manage the permissions of each lambda granularly?

JonHolman: Yes, we can do that.  The reason I did it this way here is so we don't give a function that doesn't need it, read/write access to buckets and tables.  Principle of least privilege.

peoplespete (approved):
🍻

PR #1055, opened 2025-09-16
gsBs ts precommit update

bangbay-bluetiger:
### Related ticket(s)

CMDCT-

---

### Notes

---

### Pre-review checklist

---

#### Review

#### Security

---


cli/commands/delete-topics.ts
```diff
@@ -1,7 +1,7 @@
 import { Argv } from "yargs";
 import { InvokeCommand, LambdaClient } from "@aws-sdk/client-lambda";
-import { project, region } from "../lib/consts";
-import { getCloudFormationStackOutputValues } from "../lib/utils";
+import { project, region } from "../lib/consts.js";
+import { getCloudFormationStackOutputValues } from "../lib/utils.js";
```
gmrabian, line 4: could we change our resolution so we don't need to add `.js`? seems like a downgrade considering the rest of our typescript files don't need this resolution

bangbay-bluetiger: Let me play around with that.

JonHolman: I am curious what setting(s) causes that, when you find it please let me know.  Or let me know if you would like me to hunt for it.

gmrabian: Is there a reason this folder needs a different tsconfig than our other folders? Meaning, could we use the same `root` or `ui-src` tsconfig options in this folder so that we have similar functionality across the repo?

bangbay-bluetiger: Each has its own compiling needs. `ui-src` compiles for the browser, while `app-api` and `cli` compile for Node. Maybe we could try combining the `tsconfig` for `app-api` and `cli` but I think it would involve significant refactoring.

gmrabian (approved):

PR #1069, opened 2025-09-25
Fix tar-fs vulnerability

ntsummers1:
### Description

[Dependebot has flagged a sec-vulnerability](https://github.com/Enterprise-CMCS/macpro-mdct-carts/security/dependabot/950) and we've get a fix in the same day! This is specifically the tar-fs package which is a sub-package for pa11y. Interestingly pa11y did launch an update iteself about 50 minutes ago, however, they upgraded their version of tar to a still affected one. Soooooo we'll just force it to be the patched version for now :).

### How to test

pa11y is used by our testing suite. So if the tests pass, we pass!

### Notes

pa11y: 9.0.0 -> 9.0.1
tar-fs: forced 3.1.1

---

### Pre-review checklist

---


services/ui-src/package.json
```diff
@@ -64,15 +64,14 @@
...
     "jest-launchdarkly-mock": "^2.0.3",
-    "pa11y": "^9.0.0",
+    "pa11y": "^9.0.1",
     "patch-package": "^6.4.7",
     "postinstall-postinstall": "^2.1.0",
     "react-helmet": "^6.1.0",
     "typescript": "^4.9.5"
   },
   "resolutions": {
-    "@types/react": "17.0.45",
```
bangbay-bluetiger, line 74: I'm not sure these are removable just yet.

services/ui-src/package.json
```diff
@@ -64,15 +64,14 @@
...
-    "@types/react-dom": "17.0.17"
+    "tar-fs": "3.1.1"
```
bangbay-bluetiger, line 74: This can be left off.

services/ui-src/yarn.lock
```diff
@@ -8883,10 +8880,10 @@ tabbable@^6.2.0:
   resolved "https://registry.yarnpkg.com/tabbable/-/tabbable-6.2.0.tgz#732fb62bc0175cfcec257330be187dcfba1f3b97"
   integrity sha512-Cat63mxsVJlzYvN51JmVXIgNoUokrIaT2zLclCXjRd8boZ0004U4KCs/sToJ75C6sdlByWxpYnb5Boif1VSFew==
 
-tar-fs@^3.0.8:
-  version "3.0.9"
-  resolved "https://registry.yarnpkg.com/tar-fs/-/tar-fs-3.0.9.tgz#d570793c6370d7078926c41fa422891566a0b617"
-  integrity sha512-XF4w9Xp+ZQgifKakjZYmFdkLoSWd34VGKcsTCwlNWM7QG3ZbaxnTsaBwnjFZqHRf/rROxaR8rXnbtwdvaDI+lA==
+tar-fs@3.1.1, tar-fs@^3.0.8:
```
bangbay-bluetiger, line 8883: After removing the resolution in `package.json`, `tar-fs@^3.0.8` should still be using `version "3.1.1"`.

services/ui-src/package.json
bangbay-bluetiger: Resolutions should be if we want to override what another package has declared. So, if there's a package that wants to use `tar-fs@4`, this would forcing it to use `3.1.1`.

ntsummers1: I dont disagree, but to me thats part of the dependency update process. We actively want to override what another package has declared in this instance. tar-fs will most likely update sometime in the future, as will pa11y. But when we update our dependencies, we should be mindful of the resolutions section and remove it if needed. For now, tar-fs has multiple versions with a vulnerability, and locking it to 3.1.1 ensures we're protected and calls it out in a documented fashion.

bangbay-bluetiger: I think it's going to get lost with time why this was added to `resolutions`, especially when it doesn't cause any clear build errors. But as a compromise, we can add `tar-fs` to the top-level dependencies here as in MCR.

bangbay-bluetiger: No, putting it in as top-level dependency will have the same effect on `yarn.lock` as you have now. Both `tar-fs@3.1.1` (from our `package.json`) and `tar-fs@^3.0.8` (transitive dependency for `puppeteer` are going to load version `3.1.1`.

ntsummers1: I dont think it does. Its very possible I'm mistaken in whats happenening but heres what I see when I do that:

```
"dependencies": {
    "@cmsgov/design-system": "^5.0.2",
    ...
    "sass": "^1.92.1",
    "tar-fs": "3.1.1",
    "typescript": "^5.8.3",
    ...
  },
```

Gives me:

```
tar-fs@3.1.1:
  version "3.1.1"
  resolved "https://registry.yarnpkg.com/tar-fs/-/tar-fs-3.1.1.tgz
  dependencies:
    pump "^3.0.0"
    tar-stream "^3.1.5"
  optionalDependencies:
    bare-fs "^4.0.1"
    bare-path "^3.0.0"

tar-fs@^3.1.0:
  version "3.1.0"
  resolved "https://registry.yarnpkg.com/tar-fs/-/tar-fs-3.1.0.tgz
  dependencies:
    pump "^3.0.0"
    tar-stream "^3.1.5"
  optionalDependencies:
    bare-fs "^4.0.1"
    bare-path "^3.0.0"
```

and puppeteers becomes:

```
"@puppeteer/browsers@2.10.10":
  version "2.10.10"
  resolved "https://registry.yarnpkg.com/@puppeteer/browsers/-/browsers-2.10.10.tgz
  dependencies:
    debug "^4.4.3"
    extract-zip "^2.0.1"
    progress "^2.0.3"
    proxy-agent "^6.5.0"
    semver "^7.7.2"
    tar-fs "^3.1.0"
    yargs "^17.7.2"
```

I read that as tar-fs still using the affected version :/

bangbay-bluetiger: I'm going back to not wanting `tar-fs` in `package.json`. The update in the Dependabot PR: https://github.com/Enterprise-CMCS/macpro-mdct-mfp/pull/1063 is enough to address the security issue. The PR didn't merge because of unrelated CI/CD issues.

ntsummers1: This does but it only works like that because of the resolution. Had that not happened first, this would not work which I think gets back to the fundamental difference here. Explicitly calling out this out in the resolutions section to me is exactly what that section is specifically designed for. I think if we're worried about missing it and not removing it in the future, we should make sure to update our dependency updating process instead, or leave comments as nice reminders in the future.

Overall though, I think what needs to happen is we need to get this in and fix our vulnerabilities. I'll move this to dev-dep like you mention, and make ticket to resolve this going forward so we make sure that when we update pa11y in the future we make sure to remove it once its not using an infected version. With your approval, of course.

PR #1395, opened 2026-04-07
CMDCT-5588 + CMDCT-5635: [WP Initiatives] Key Metrics Table (I)

karla-vm:
### Description

This PR covers the following functionality of the new `Work Plan` initiatives key metrics table:
- Blank state
- Adding new key metrics
- Deleting key metrics

As a follow-up, these need to be addressed:
- Editing an existing key metric
- Confirmation modal prior to key metric deletion
- Reloading and hydration behavior upon adding and deleting key metrics

### Related ticket(s)

CMDCT-5588
CMDCT-5635

---
### How to test

- Log in as a state user
- Create a Work Plan
- Navigate to `/wp/state-or-territory-specific-initiatives/initiatives`
- Create an initiative
- Select **Edit**
- Scroll down the page to the `Key Metrics` section

The table blank state should look like this:

Screenshot 2026-04-16 at 4 07 44 PM

- Add a new key metric

You should something like the following:

Screenshot 2026-04-16 at 4 00 48 PM

**NOTE:** The first time you add a key metric, it won't show up _yet_, you need to return to the initiatives table by clicking on "Previous" to close the overlay and then it will load. This will be addressed in a follow-up PR.

### Notes

N/A

---
### Pre-review checklist

---
#### Review

#### Security


services/ui-src/src/components/tables/DynamicTableRows.tsx
```diff
@@ -41,7 +43,11 @@ export const DynamicTableRows = ({
 
   // Add rows from fieldData
   useEffect(() => {
-    const rows = localFieldData?.[dynamicRowsTemplate.id];
+    // if there is an entity type "Initiatives", handle Key Metrics
+    const rows =
```
bangbay-bluetiger, line 47: When I create a new initiative, `defineInitiative_keyMetrics_performanceIndicators` doesn't exist on `formData` but it exists in `localFieldData`. I think we need to use a lookup within `localFieldData` instead of using `formData` straight up.

https://github.com/user-attachments/assets/2b3ff2a3-eebe-446f-8a51-c3121cd9c45e

services/ui-src/src/components/tables/DynamicTableRows.tsx
```diff
@@ -102,47 +112,51 @@ export const DynamicTableRows = ({
...
+                      type="button"
+                      variant={"unstyled"}
+                    >
+                      <Image src={cancelIcon} alt={`Delete ${name}`} />
+                    </Button>
+                  )}
+                </Flex>
+              </Td>
+            </Tr>
+            {/* TODO: DeleteEntityModal */}
```
bangbay-bluetiger, line 158: If we add a confirm modal, I think it should live in the `EntityModalTable` and then get called within here with a new prop like `openConfirmModal`.

services/app-api/forms/routes/wp/flags/wpSarRelease2025/state-or-territory-specific-initiatives/initiatives.ts
```diff
@@ -77,7 +77,8 @@ const keyMetricsDynamicFieldsToReturn = [
               validation: {
                 type: ValidationType.TEXT,
                 nested: true,
-                parentFieldName: `${keyMetricsDynamicRowId}-dataSource`,
+                parentFieldName: `dataSource`,
```
bangbay-bluetiger, line 80: I think this should stay `${keyMetricsDynamicRowId}-dataSource`. The full name of the parent field is `defineInitiative_keyMetrics_performanceIndicators-dataSource`.

services/ui-src/src/components/tables/DynamicTableProvider.tsx
```diff
@@ -203,12 +205,33 @@ export const DynamicTableProvider = ({ children }: any) => {
       const { dynamicFieldId, dynamicTemplateId, fieldType } = getFieldParts(
         hydratedField.id
       );
-      const templateFieldData = localFieldData?.[dynamicTemplateId] || [];
+
+      const templateFieldData =
+        // if there is an entity type "Initiatives", handle Key Metrics
+        entityType === entityTypes[0]
```
ntsummers1, line 211: Would love to see an enum that checks for this rather than an array call here, something like:

```
export enum EntityType {
  INITIATIVES = "initiatives",
  TARGETPOPULATIONS = "target_populations",
  ...
}
```

services/ui-src/src/components/tables/DynamicTableProvider.tsx
```diff
@@ -203,12 +205,33 @@ export const DynamicTableProvider = ({ children }: any) => {
...
         (field: DynamicFieldShape) => field.id === dynamicFieldId
       );
 
       hydrateValue = currentField?.[fieldType];
+
+      // handle Key Metrics table edge cases
+      if (entityType === entityTypes[0]) {
+        switch (fieldType) {
+          case "dataSource":
+            currentField?.dataSource[0].value === "Other, specify"
```
ntsummers1, line 225: Is is guaranteed that datasource will have something in the 0 index? If not we should guard against it.

services/ui-src/src/components/tables/DynamicTableProvider.tsx
```diff
@@ -419,13 +442,19 @@ export const DynamicTableProvider = ({ children }: any) => {
   const removeDynamicRow = async (
     dynamicTemplateId: string,
     dynamicFieldId: string,
+    entityType?: string,
     updatedFields: FieldInfo[] = []
   ) => {
-    const rows = localFieldData?.[dynamicTemplateId] || [];
+    const rows =
+      entityType === entityTypes[0]
```
ntsummers1, line 449: Same guard note as above

services/ui-src/src/components/tables/DynamicTableProvider.tsx
```diff
@@ -499,6 +528,7 @@ interface DisplayCellOptions {
   cell: string | FormField;
   columnId: string;
   disabled: boolean;
+  entityType?: string;
```
ntsummers1, line 531: This can by type `EntityType`

services/ui-src/src/components/tables/DynamicTableRows.tsx
```diff
@@ -102,47 +112,51 @@ export const DynamicTableRows = ({
...
+                  {!disabled && (
+                    <Button
+                      onClick={() =>
+                        removeDynamicRow(
+                          dynamicRowsTemplate.id,
+                          dynamicId,
+                          entityType,
+                          updatedFieldsCallback(dynamicId, localFieldData)
+                        )
+                      }
```
ntsummers1, line 147: I was unable to get this to actually work. Leaving the page and coming back showed it still there as well

ntsummers1 (commented):
Looks like this is coming along! A couple call outs and typing nits

services/ui-src/src/components/tables/DynamicTableProvider.tsx
karla-vm: Great call, I'll update this! 💯

services/ui-src/src/components/tables/DynamicTableRows.tsx
karla-vm: Not sure what's happening there, I just checked on my end and it's working... 🤔

services/ui-src/src/components/tables/DynamicTableProvider.tsx
karla-vm: Yes, it's a required option field, so it should always have a key-value pair in the zero index. I'll add a `?` to guard against `undefined` values though!

services/ui-src/src/components/tables/DynamicTableRows.tsx
karla-vm: Hi! That attribute actually doesn't exist in `localFieldData` in the case of initiatives, I just checked on the report object. Were you seeing it on the `localFieldData` object?

ajaitasaini (dismissed):
tested and looks great!!

services/ui-src/src/components/tables/DynamicTableRows.tsx
bangbay-bluetiger: I'm finding that this is still not updating the table when I create a key metric for the first time. It needs a look up, something like this:
```
const initiativeData = (localFieldData?.initiative || [])
  .find((t: DynamicFieldShape) => t.id === formData?.id);

const rows =
  entityType === EntityType.INITIATIVE
    ? initiativeData?.[dynamicRowsTemplate.id]
    : localFieldData?.[dynamicRowsTemplate.id];
```

Plus it needs a corresponding change in the `DynamicTableProvider`.

services/ui-src/src/components/tables/DynamicTableProvider.tsx
```diff
@@ -207,12 +209,33 @@ export const DynamicTableProvider = ({ children }: any) => {
       const { dynamicFieldId, dynamicTemplateId, fieldType } = getFieldParts(
         hydratedField.id
       );
-      const templateFieldData = localFieldData?.[dynamicTemplateId] || [];
+
+      const templateFieldData =
```
bangbay-bluetiger: I would change this to include a lookup, like:
```
const initiativeData = (localFieldData?.initiative || [])
  .find((t: DynamicFieldShape) => t.id === formData?.id);

const templateFieldData =
  // if there is an entity type "Initiatives", handle Key Metrics
  entityType === EntityType.INITIATIVE
    ? initiativeData?.[dynamicTemplateId]
    : localFieldData?.[dynamicTemplateId];

const currentField = (templateFieldData || []).find(
  (field: DynamicFieldShape) => field.id === dynamicFieldId
);
```

services/ui-src/src/components/tables/DynamicTableProvider.tsx
```diff
@@ -430,13 +453,19 @@ export const DynamicTableProvider = ({ children }: any) => {
   const removeDynamicRow = async (
     dynamicTemplateId: string,
     dynamicFieldId: string,
+    entityType?: string,
     updatedFields: FieldInfo[] = []
   ) => {
-    const rows = localFieldData?.[dynamicTemplateId] || [];
+    const rows =
```
bangbay-bluetiger: Need a lookup here, too, like:
```
const initiativeData = (localFieldData?.initiative || [])
  .find((t: DynamicFieldShape) => t.id === dynamicFieldId);

const rows =
  entityType === EntityType.INITIATIVE
    ? initiativeData?.[dynamicTemplateId]
    : localFieldData?.[dynamicTemplateId];

// Remove row to be deleted
const updatedRows = (rows || []).filter((row: DynamicFieldShape) => {
  return row.id !== dynamicFieldId;
});
```

services/ui-src/src/components/tables/DynamicTableRows.tsx
```diff
@@ -83,16 +89,20 @@ export const DynamicTableRows = ({
...
-          <Tr
-            key={dynamicId}
-            id={dynamicId}
-            ref={(el) => {
-              rowRefs.current[rowIndex] = el;
-            }}
-          >
-            {dynamicRowsTemplate.props?.dynamicFields.map(
-              (field: FormField, cellIndex: number) => (
+          <>
```
bangbay-bluetiger: Having this `<>` wrapper is causing this key error because keys need to be on the parent element:

Image

bangbay-bluetiger (approved):

PR #1398, opened 2026-04-10
FRF - Build the PDF for FRF Form with tables

ajaitasaini:
### Description

Adds tables within FRF PDF

### Related ticket(s)

[CMDCT-5666](https://jiraent.cms.gov/browse/CMDCT-5666)

---
### How to test

https://d18dguw9dv9wmu.cloudfront.net/

Head to Review PDF for an ABCD form, confirm that the tables are showing correctly like in the Figma: https://www.figma.com/design/cNFWWAgpTPfir1kglCnOJ2/MFP-FRF--Internal-?node-id=5932-7239&t=BqiqCrBY4Adf1YeC-0

### Notes

---
### Pre-review checklist

---
#### Review

#### Security


services/ui-src/src/utils/export/exportFieldTableHelpers.tsx
```diff
@@ -0,0 +1,212 @@
...
+      return table.dynamicRowsTemplate.props.dynamicFields.map((f: any) => {
+        const { fieldType } = getFieldParts(f.id);
+
+        // This puts Misc Costs: or Other: before the value
+        const dynamicLabel = f.props.dynamicLabel;
+        let fieldValue = item[fieldType];
+
+        // Number needs to be masked
+        if (f.type === ReportFormFieldType.NUMBER) {
+          fieldValue = maskResponseData(fieldValue, f.props.mask);
```
bangbay-bluetiger, line 111: Ran into issue where I added a Sub Recipient without a percentage override, and the render failed on this line because `maskResponseData` doesn't take a null `fieldValue`. Could need to just change the conditional to `if (f.type === ReportFormFieldType.NUMBER && fieldValue)`.

services/ui-src/src/utils/export/exportFieldTableHelpers.tsx
```diff
@@ -0,0 +1,212 @@
+import { ReactElement } from "react";
```
bangbay-bluetiger, line 1: I think this file would be better under the `services/ui-src/src/utils` folder. There's already a `services/ui-src/src/utils/other/export.tsx` file there, but I think adding a new `export` folder and dropping this file in there would be better.

bangbay-bluetiger (changes requested):
Looks good to me! Just the issue with the null `fieldValue`. Also, please add some tests for the helpers. 😀

services/ui-src/src/components/tables/Table.tsx
```diff
@@ -59,44 +59,44 @@ export const Table = ({
...
+            {row.map((cell: string, rowIndex: number) => (
+              <Td
+                key={`${cell}${rowIndex}-body-cell`}
+                sx={{
+                  tableCell: border ? sx.tableCellBorder : sx.tableCell,
+                  color: cell == notAnsweredText ? "error_darker" : "",
+                }}
+                aria-label={ariaOverride?.bodyRows?.[index][rowIndex]}
+              >
+                {isValidElement(cell) ? cell : sanitizeAndParseHtml(cell)}
```
bangbay-bluetiger, line 73: What does `isValidElement` do?

services/ui-src/src/utils/export/exportFieldTableHelpers.test.tsx
```diff
@@ -0,0 +1,216 @@
...
+jest.mock("../../components/tables/Table", () => ({
+  __esModule: true,
+  Table: jest.fn((props) => <div data-testid={props["data-testid"]} />),
+}));
+
+jest.mock("utils/other/parsing", () => ({
+  parseCustomHtml: (html: string) => html,
+}));
+
+jest.mock("utils/other/mask", () => ({
```
bangbay-bluetiger: Why mock out the `Table` component and these utils?

services/ui-src/src/utils/export/exportFieldTableHelpers.test.tsx
```diff
@@ -0,0 +1,216 @@
...
+    });
+  });
+
+  describe("renderCalculationTables()", () => {
+    beforeEach(() => {
+      mockedUseStore.mockReturnValue({
+        report: { fieldData },
+      } as any);
+    });
+    it("should render calculation tables", async () => {
```
bangbay-bluetiger: Could get a smidgen more test coverage by testing for more sections and percentages. I'd do something like this:

```
const tables = [
  {
    id: "qualifiedHcbs",
    name: "Qualified HCBS",
    pct: 50,
  },
  {
    id: "demonstrationServices",
    name: "Demonstration Services",
    pct: 75,
  },
  {
    id: "other",
    name: "Other",
    pct: 0,
  }
];

tables.forEach(({ id, pct, name }) => {
  render(<>{renderCalculationTables(section(id, name), fieldData(id, pct), 0)}</>);
  expect(screen.getByTestId(`service-table-${id}`)).toBeVisible();
  expect(screen.getByText(`Percentage: ${pct}%`)).toBeVisible();
});
```
And alter `section` and `fieldData` to be functions:
```
const fieldData = (id, pct) => ({
  [`fmap_${id}Percentage`]: pct,
  ...etc.
});

const section = (id, name) => ({
  name,
  form: {
    tables: [
      {
        id,
  ...etc.
});

```

services/ui-src/src/utils/export/exportFieldTableHelpers.test.tsx
```diff
@@ -0,0 +1,216 @@
...
+      },
+    };
+
+    beforeEach(() => {
+      mockedUseStore.mockReturnValue({
+        report: mockReport,
+      } as any);
+    });
+
+    it("should render service table body correctly", () => {
```
bangbay-bluetiger: Prefer to use `test` for test cases instead of `it` for consistency.

bangbay-bluetiger (changes requested):
The tables look good to me! Just want to get the test coverage level up to pass our checks.

services/ui-src/src/components/tables/Table.tsx
ajaitasaini: I was using `isValidElement `because the cell is now of type string or ReactNode (had to do this for rendering purposes), and isValidElement was a safety check on the cell. Let me know if you think there's an easier way to do this!

services/ui-src/src/utils/export/exportFieldTableHelpers.test.tsx
ajaitasaini: this was some old code i forgot to remove, thanks for the catch

services/ui-src/src/components/tables/Table.tsx
bangbay-bluetiger: I think it's fine to use `isValidElement`. Could also switch the check around to:
```
{typeof cell === "string" ? sanitizeAndParseHtml(cell) : cell}
```

If `cell` isn't a string, it'll affect how the ``key={`${cell}${rowIndex}-body-cell`}`` is set, so that would need to do something else there. Also, the type should be changed to `cell: string | ReactNode`.

bangbay-bluetiger (approved):
Non-blocking comments. Can be followed up on later.

jessabean (approved):
Tested and can confirm things are working great! I think the suggestions thus far have been good. Thank you for carrying this across the finish line!

PR #1417, opened 2026-04-20
Kafka refactor

benmartin-coforma:
### Description
This PR refactors kafka-source-lib.ts to fit more nicely with our codebase:
- It has always stuck out as one of the only (possibly _the_ only) class defined in any of the MDCT apps.
- It has been copy-pasted around between apps, and modified for each, so that it _mostly but not quite_ matches, and thus cannot be synced.
- It isolates app-specific information (such as table names) in an extreme way: these values are pulled out to postKafkaData.ts, without the context of exactly how they are used.

The replacement is kafkaLib.ts:
- It uses the same `export const handler = createHandler(callbacks)` pattern as the rest of our API endpoints.
- The callbacks allow each app to customize postKafkaData.ts as needed, in theory allowing kafkaLib.ts to be synced.
- The callbacks contain all app-specific information in context, making it clear how (for example) Dynamo table names correspond to complete Kafka topic names.
- All interaction with `process.env` is contained within the app code.
- The Kafka producer management and connection management code is more self-contained; see `LazyProducer` in kafkaLib.ts.
- The new tests completely cover postKafkaData.ts and kafkaLib.ts in conjunction; we're not just testing the library but also our use of the library.

I created this PR on MFP because it has one of our most complicated Kafka integrations. If the team approves, I will create new PRs for every repo.

### Related ticket(s)
CMDCT-5269

---
### How to test
https://d363fhxuvfieqn.cloudfront.net/

Mostly, testing this will mean creating & modifying a few reports, then scrutinizing the logs in the deployed environment.

---
### Pre-review checklist

---
#### Review

#### Security


services/app-api/handlers/kafka/kafkaLib.ts
```diff
@@ -0,0 +1,174 @@
...
+      return;
+    }
+
+    const batch = { topicMessages };
+    console.debug(`Outgoing message batch: ${JSON.stringify(batch, null, 2)}`);
+
+    await LazyProducer.sendBatch(config, batch);
+
+    console.info(`Processing complete.`);
+  };
```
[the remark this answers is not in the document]
benmartin-coforma: Reorganized.

services/app-api/handlers/kafka/postKafkaData.ts
```diff
@@ -0,0 +1,111 @@
...
+    },
+    ssl: {
+      rejectUnauthorized: false,
+    },
+  };
+};
+
+const getDynamoTopic: GetDynamoTopic = (record) => {
+  const namespace = process.env.topicNamespace ?? "";
+  const tables = ["wp-reports", "sar-reports"];
```
gmrabian, line 41: question for app team: do we need the new expenditures/FRF table here?

services/app-api/handlers/kafka/postKafkaData.ts
```diff
@@ -0,0 +1,111 @@
...
+    console.warn(`Ignoring record: no matching table`);
+    return undefined;
+  }
+
+  return `${namespace}aws.mdct.mfp.${table}.v0`;
+};
+
+const getS3Topic: GetS3Topic = (record) => {
+  const namespace = process.env.topicNamespace ?? "";
+  const rules = [
```
gmrabian, line 56: question for app team: do we need the new expenditures/FRF bucket here?

services/app-api/handlers/kafka/kafkaLib.ts
```diff
@@ -0,0 +1,185 @@
...
+      topic,
+      messages: messages.filter((m) => m.topic === topic).map((m) => m.message),
+    }));
+
+    if (topicMessages.length === 0) {
+      console.warn(`Ignoring event: no messages to send.`);
+      return;
+    }
+
+    const batch = { topicMessages };
```
gmrabian, line 32: The object nesting now contains an extra layer. Is that a problem at all for dataconnect? Should we make sure it's 1:1 same format?

Image
Image

benmartin-coforma: I believe the format of the message we're sending is unchanged. The old version did [log something other than what it sent](https://github.com/Enterprise-CMCS/macpro-mdct-mfp/blob/8572323b4edfdc00e15d1745ec8d846799ad5a1f/services/app-api/utils/kafka/kafka-source-lib.ts#L240-L242) - if those screenshots are from lambda logs, then I think that's the source of confusion.

services/app-api/handlers/kafka/kafkaLib.ts
```diff
@@ -0,0 +1,185 @@
...
+      const key = s3.object.key;
+      const includeData = record.eventName !== "ObjectRemoved";
+      const value = includeData ? await getS3Object!(bucket, key) : "";
+      messages.push({
+        topic,
+        message: { headers, partition: 0, key, value },
+      });
+    }
+  }
+  return messages;
```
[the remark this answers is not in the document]
benmartin-coforma: Reorganized again.

services/app-api/handlers/kafka/kafkaLib.ts
```diff
@@ -0,0 +1,182 @@
...
+  } else if ("s3" in record) {
+    const topic = getS3Topic!(record);
+    if (!topic) {
+      return undefined;
+    }
+    const { eventTime, eventName, s3 } = record;
+    const headers = { eventName, eventTime };
+    const bucket = s3.bucket.name;
+    const key = s3.object.key;
+    const includeData = record.eventName !== "ObjectRemoved";
```
JonHolman:
```suggestion
    const includeData = !record.eventName.startsWith("ObjectRemoved");
```
S3 event delete notifications are `ObjectRemoved:Delete` and `ObjectRemoved:DeleteMarkerCreated`.

https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-how-to-event-types-and-destinations.html

services/app-api/handlers/kafka/kafkaLib.ts
```diff
@@ -0,0 +1,182 @@
...
+      process.removeListener(signal, disconnect);
+      process.once(signal, disconnect);
+    }
+
+    const kafka = new Kafka(config);
+    producer?.disconnect();
+    producer = kafka.producer();
+    producer.on("producer.disconnect", disconnect);
+    connected = true;
+    await producer.connect();
```
JonHolman:
```suggestion
    await producer.connect();
    connected = true;
```

services/app-api/handlers/kafka/kafkaLib.ts
benmartin-coforma: Excellent, thank you!

services/app-api/handlers/kafka/kafkaLib.ts
benmartin-coforma: I chose this ordering intentionally, because I was worried about race conditions. My reasoning is that, if a second call to `sendBatch()` is made while we wait to connect for the first call, we shouldn't initialize a second connection.

Does the JS event loop make it so that we don't have to worry about that kind of thing? Should I instead introduce a new lock variable to account for this? Given the nature of JS, _can_ I even create a lock variable, or would that too be subject to race conditions?

JonHolman: I don't think the event loop makes this safe across the await.  Setting connected before the await lets another call skip connect before the producer is ready; setting it after lets a second call start a duplicate connect.  Maybe we should store the in-flight connect() promise and have concurrent callers await it, clearing only on failure so retries still work.

benmartin-coforma: I like the shared promise idea! Implemented.

If we ever have an actual failure at any point, the entire process will exit - as before, and as designed.

services/app-api/handlers/kafka/kafkaLib.ts
gmrabian: Ah ok I see now. Thanks!

gmrabian (dismissed):
Awesome! 🎉

JonHolman (dismissed):

services/app-api/handlers/kafka/kafkaLib.ts
```diff
@@ -0,0 +1,188 @@
...
+export const kafkaHandler = (callbacks: KafkaCallbacks) => {
+  return async (event: AwsStreamEvent) => {
+    const config = callbacks.getConfig();
+    if (config === undefined) {
+      return;
+    }
+    console.debug(`Raw event: ${JSON.stringify(event, null, 2)}`);
+
+    const messages: { topic: string; message: Message }[] = [];
+    for (let record of event.Records ?? []) {
```
bangbay-bluetiger: Think this should be `const record`. I don't see `record` reassigned in the loop.

services/app-api/handlers/kafka/kafkaLib.ts
```diff
@@ -0,0 +1,188 @@
...
+
+/**
+ * Translate an AWS stream event record into a Kafka message.
+ * @returns `undefined` if a topic cannot be found.
+ */
+const convertToMessage = async (
+  record: NonNullable<AwsStreamEvent["Records"]>[number],
+  { getDynamoTopic, getS3Topic, getS3Object }: KafkaCallbacks
+) => {
+  if ("dynamodb" in record) {
```
bangbay-bluetiger, line 47: More a stylistic choice, but you could put the `getDynamoTopic/getS3Topic` switch in another function and do an early undefined return once. Something like this:
```
const getTopic = (record, getDynamoTopic, getS3Topic) => {
  if ("dynamodb" in record) {
    return getDynamoTopic?.(record);
  }

  if ("s3" in record) {
    return getS3Topic?.(record);
  }
};

const topic = getTopic(record, getDynamoTopic, getS3Topic)

if (!topic) return;
```

services/app-api/handlers/kafka/kafkaLib.ts
```diff
@@ -0,0 +1,188 @@
...
+      return undefined;
+    }
+    const { eventTime, eventName, s3 } = record;
+    const headers = { eventName, eventTime };
+    const bucket = s3.bucket.name;
+    const key = s3.object.key;
+    // May be s3:ObjectRemoved:Delete or s3:ObjectRemoved:DeleteMarkerCreated
+    const includeData = !record.eventName.includes("ObjectRemoved");
+    const value = includeData ? await getS3Object!(bucket, key) : "";
+    return { topic, message: { headers, partition: 0, key, value } };
```
bangbay-bluetiger, line 72: Since the return pattern is the same, this could live outside of the conditionals for a single return. Something like:
```
let headers;
let key;
let value;

if ("dynamodb" in record) {
  etc.
}

if ("s3" in record) {
  etc.
}

return { topic, message: { headers, partition: 0, key, value } };
```

services/app-api/handlers/kafka/kafkaLib.ts
```diff
@@ -0,0 +1,188 @@
...
+ * Connects / disconnects only as necessary.
+ * Note: config cannot be changed after the first call to sendBatch.
+ */
+const LazyProducer = (() => {
+  let producer: Producer | undefined;
+  let connectPromise: Promise<void> | undefined;
+  let connected = false;
+
+  const connect = async (config: KafkaConfig) => {
+    for (let signal of ["SIGTERM", "SIGINT", "SIGUSR2", "beforeExit"]) {
```
bangbay-bluetiger: Can be `const signal` since it isn't reassigned.

services/app-api/handlers/kafka/kafkaLib.ts
```diff
@@ -0,0 +1,188 @@
...
+      process.removeListener(signal, disconnect);
+      process.once(signal, disconnect);
+    }
+
+    const kafka = new Kafka(config);
+    producer?.disconnect();
```
bangbay-bluetiger: Does this need an `await`?

services/app-api/handlers/kafka/kafkaLib.ts
```diff
@@ -0,0 +1,188 @@
...
+    producer = kafka.producer();
+    producer.on("producer.disconnect", disconnect);
+    // If sendBatch is called twice before we can finish connecting once,
+    // both calls will wait for the same connection promise.
+    connectPromise ??= (async () => {
+      producer.connect();
```
bangbay-bluetiger: Does this need an `await`?

services/app-api/handlers/kafka/postKafkaData.ts
bangbay-bluetiger: It'll be added in a future task.

services/app-api/handlers/kafka/postKafkaData.ts
```diff
@@ -0,0 +1,111 @@
...
+    {
+      bucket: reportBuckets.WP,
+      folder: "fieldData/",
+      topic: "wp-form",
```
bangbay-bluetiger, line 60: Is there a particular reason to use the string value here instead of the constants?

services/app-api/handlers/kafka/postKafkaData.ts
jessabean: Soon: https://jiraent.cms.gov/browse/CMDCT-4813

services/app-api/handlers/kafka/kafkaLib.ts
benmartin-coforma: That's a good point! But the tradeoff to have just one "if no topic" check would be to have multiple "if dynamo/s3" checks. I tried it out but I think I like this better.

services/app-api/handlers/kafka/kafkaLib.ts
benmartin-coforma: Yes, but structuring the code to re-use the one return statement would mean increasing the distance between variable declaration + assignment + use for those pieces of the message, without saving any lines of code.

services/app-api/handlers/kafka/postKafkaData.ts
benmartin-coforma: Yes - one of the driving motivations behind this whole refactor was to make it easier to answer the question "what topic do we send this message on?". That's why I was so intent on moving the topic namespace, prefix, version, and selection logic all into one function (per event source). Using a constant would mean requiring a reader to jump to another file (or at least move their mouse & remember what they saw) to learn the value of this piece.

This isn't my strongest opinion, though; if you object I can re-add the constant references. If you don't, I'll push another update removing the constants altogether; postKafkaData was the only code that referenced them.

bangbay-bluetiger: Sounds reasonable to me. We can remove the constants.

bangbay-bluetiger (approved):

jessabean (approved):
Let's test the heck out of this in val and sync with DataConnect to make sure they're not seeing anything unexpected. Nice job! 🚀

BearHanded (approved):

PR #1418, opened 2026-04-21
Cmdct 5786 - aria labels for user created rows should be saved with specific names and categories

britt-mo:
### Description

Per the card:
MFP table row delete actions incorrect label

"The label should say Delete <row name>. In the above picture. If the row name is Other: testing then the delete label should be Delete Other: testing, not Delete 18cf4eb-08d-7e8b-0e1c-f13ca660da0."

The user-created rows won't get aria labels autosaved with a UUID attached. Rows with category names (Supplemental Services, Miscellaneous Costs) will include the row name in the aria-label. Otherwise it will say "Delete [name]", like in Personnel 

### Related ticket(s)

CMDCT-5786

---
### How to test
Create a financial report and check that after you create rows, their aria-labels are name/category specific.

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

### DIFF
--- a/services/ui-src/src/components/tables/DynamicTableRows.tsx
+++ b/services/ui-src/src/components/tables/DynamicTableRows.tsx
@@ -8,7 +8,10 @@ import {
   DynamicFieldShape,
   DynamicRowsTemplate,
   FormField,
+  ReportType,
 } from "types";
+// utils
+import { useStore } from "utils";
 // assets
 import cancelIcon from "assets/icons/icon_cancel_x_circle.png";
 
@@ -30,6 +33,8 @@ export const DynamicTableRows = ({
     localFieldData,
     removeDynamicRow,
   } = useContext(DynamicTableContext);
+  const { report } = useStore();
+  const isFinancialReport = report?.reportType === ReportType.FINANCIAL_REPORT;
   const dynamicLabel = dynamicRowsTemplate.props?.dynamicFields.find(
     (field: FormField) => field.props?.dynamicLabel
   )?.props?.dynamicLabel;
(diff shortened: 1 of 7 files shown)


services/ui-src/src/components/tables/DynamicTableRows.tsx
```diff
@@ -81,7 +89,15 @@ export const DynamicTableRows = ({
...
-        const name = row.name || dynamicId;
+        const name = row.category || row.title || row.name || dynamicId;
+        const editLabel = `Edit ${name}`;
+        let deleteLabel = `Delete ${name}`;
+
+        if (isFinancialReport && dynamicLabel === "Misc. Costs:") {
+          deleteLabel = `Delete Misc. Costs: ${name}`;
+        } else if (isFinancialReport && dynamicLabel === "Other:") {
+          deleteLabel = `Delete Other: ${name}`;
+        }
```
britt-mo: @cy-by in the card you specified that if the row was titled `Other`, then the label should include it like so: `Other: title`. So I applied that to `Misc. Costs` as well because that section is styled similarly. But Personnel isn't labeled like those, so it remains as `Delete: [name]`.

services/ui-src/src/components/tables/DynamicTableRows.tsx
```diff
@@ -81,7 +89,15 @@ export const DynamicTableRows = ({
 
       {localDynamicRows.map((row, rowIndex: number) => {
         const dynamicId = row.id;
-        const name = row.name || dynamicId;
+        const name = row.category || row.title || row.name || dynamicId;
+        const editLabel = `Edit ${name}`;
+        let deleteLabel = `Delete ${name}`;
+
+        if (isFinancialReport && dynamicLabel === "Misc. Costs:") {
```
bangbay-bluetiger: I don't think it's necessary to check what kind of report this is.  You can also use `dynamicLabel` directly in `deleteLabel`, something like:
```
const deleteLabel = ["Delete", dynamicLabel, name].filter(Boolean).join(" ");
```

bangbay-bluetiger (changes requested):

cy-by (dismissed):
Delete buttons have dynamic labeling now 👍🏻 

Image

Image

Image

Image

bangbay-bluetiger (approved):

PR #1444, opened 2026-05-13
CMDCT-5875:  SAR Initiatives Key Metrics table

karla-vm:
### Description

Update the SAR initiatives `Key Metrics` table, which should auto-populate with initiative data from its corresponding Work Plan.

### Related ticket(s)

CMDCT-5875
CMDCT-5971

---
### How to test

**The hard way**
- Log in as a state user
- Create, fill, and submit a Work Plan, making sure to add at least one (1) key metric
- Log in as an admin
- Approve that Work Plan
- Create a SAR
- Navigate to `/sar/state-or-territory-specific-initiatives`
- Enter an initiative

**The less hard way**
- Start the app locally `./run local`
- Open a new terminal window and run `yarn db:seed`
- Create an approved Work Plan
- Log in as a state user
- Create a SAR
- Navigate to `/sar/state-or-territory-specific-initiatives`
- Enter an initiative

**Final step for either method**
Verify that the Key Metrics table is auto-populated with the values you entered in the WP and you can enter a value in the `Actual (optional)` field ✨ 

Screenshot 2026-05-14 at 10 24 13 AM

### Notes

N/A

---
### Pre-review checklist

---
#### Review

#### Security

### DIFF
--- a/services/ui-src/src/components/reports/EntityProvider.tsx
+++ b/services/ui-src/src/components/reports/EntityProvider.tsx
@@ -34,8 +34,10 @@ export const EntityProvider = ({ children }: EntityProviderProps) => {
    */
   const prepareEntityPayload = (updateData: AnyObject) => {
     const entityType = selectedEntity!.type;
+    const currentEntities = selectedEntity
+      ? { ...report?.fieldData?.[entityType], ...selectedEntity }
+      : report?.fieldData?.[entityType];
 
-    const currentEntities = report?.fieldData?.[entityType];
     const selectedEntityIndex = currentEntities?.findIndex(
       (x: EntityShape) => x.id === selectedEntity?.id
     );
@@ -53,7 +55,7 @@ export const EntityProvider = ({ children }: EntityProviderProps) => {
     () => ({
       prepareEntityPayload,
     }),
-    [selectedEntity, report]
+    [selectedEntity]
   );
 
   return (
--- a/services/ui-src/src/components/reports/EntityProvider.tsx
+++ b/services/ui-src/src/components/reports/EntityProvider.tsx
@@ -34,6 +34,7 @@ export const EntityProvider = ({ children }: EntityProviderProps) => {
    */
   const prepareEntityPayload = (updateData: AnyObject) => {
     const entityType = selectedEntity!.type;
+
     const currentEntities = report?.fieldData?.[entityType];
     const selectedEntityIndex = currentEntities?.findIndex(
       (x: EntityShape) => x.id === selectedEntity?.id
@@ -52,7 +53,7 @@ export const EntityProvider = ({ children }: EntityProviderProps) => {
     () => ({
       prepareEntityPayload,
     }),
-    [selectedEntity]
+    [selectedEntity, report]
   );
 
   return (
--- a/services/ui-src/src/components/reports/EntityProvider.tsx
+++ b/services/ui-src/src/components/reports/EntityProvider.tsx
@@ -34,10 +34,7 @@ export const EntityProvider = ({ children }: EntityProviderProps) => {
    */
   const prepareEntityPayload = (updateData: AnyObject) => {
     const entityType = selectedEntity!.type;
-    const currentEntities = selectedEntity
-      ? { ...report?.fieldData?.[entityType], ...selectedEntity }
-      : report?.fieldData?.[entityType];
-
+    const currentEntities = report?.fieldData?.[entityType];
     const selectedEntityIndex = currentEntities?.findIndex(
       (x: EntityShape) => x.id === selectedEntity?.id
     );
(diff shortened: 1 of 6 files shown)


britt-mo (commented):
I know the `Actual` field is `optional`, but I was wondering if there's a need for validation, or do we not care if the user adds numbers or text?

karla-vm:
> I know the `Actual` field is `optional`, but I was wondering if there's a need for validation, or do we not care if the user adds numbers or text?

I _think_ there's no need for validation in that sense, I'll run it by Anna!

services/ui-src/src/components/reports/EntityProvider.tsx
```diff
@@ -34,10 +34,7 @@ export const EntityProvider = ({ children }: EntityProviderProps) => {
    */
   const prepareEntityPayload = (updateData: AnyObject) => {
     const entityType = selectedEntity!.type;
-    const currentEntities = selectedEntity
-      ? { ...report?.fieldData?.[entityType], ...selectedEntity }
-      : report?.fieldData?.[entityType];
-
```
britt-mo, line 40: Can you elaborate on why this change needed to be undone? Just for my own understanding

karla-vm: Good question! So, @ntsummers1 pointed out that if we just add the `report` to the EntityProvider memo value [here](https://github.com/Enterprise-CMCS/macpro-mdct-mfp/blob/779174101bc2b42d5c25da50d766e0fc599ae396/services/ui-src/src/components/reports/EntityProvider.tsx#L55), we actually don't need to spread the selected entity details to the entities. I was actually going to ask Nick if we could discuss that later on because I'd love to understand the use of that hook a bit better myself as well 😄

ntsummers1: Ok so first things first, whats the difference between useMemo and useEffect:

**TL;DR:** Use useMemo when you want to calculate a value based on props or state without re-running the logic on every render. Use useEffect when you want to take an action that interacts with something outside of the component. 

useMemo has a couple nice things going for it: “Memoizing” a value to avoid expensive recalculations. Anytime you hear the word memo or memoizing I want you to think “Memory.” We’re storing a value in memory for a quick look up.  It also will crucially return a value for you every time a component renders.

useEffect on the other hand returns nothing on its render, and is usually used for doing things like API calls, subscriptions, manual DOM changes. These run after rendering has been committed to the screen.

They both look similar but you can now see the differences here:
```
const providerValue = useMemo(
    () => ({
      prepareEntityPayload,
    }),
    [selectedEntity, report]
  );

useEffect(() => {
    fireTealiumPageView(
      user,
      window.location.href,
      pathname,
      isApparentReportPage(pathname)
    );
  }, [key]);
```
You can see with useMemo we’re returning a value called providerValue that we calculate from the function prepareEntityPayload. With useEffect we’re making an api call and crucially not returning anything

The last thing to know is what the values mean in the array. We call this the dependency array. useMemo has `selectedEntity` and `report` in its case. useEffect has `key`. They both, however, are doing the same thing: When ever the value stored in the array is changed, run that useMemo or useEffect again. So if selectedEntity changes? Run useMemo. The key changes? Run the useEffect.

—--------

**Actual Chat About The Fix:**

Great, that's all well and good, but why does report being in this useMemo dependency array matter? Well, lets look at the useMemo function…

prepareEntityPayload interacts directly with the report from the store. But because report wasn’t in the dependency array, the memoized function (Think function we’ve stored and saved the value of in memory!) only got a fresh report when selectedEntity changed. This occurres when the overlay was first opened.

Where we can lose data though is through this scenario of events:

- User opens initiative → prepareEntityPayload captures the stored report (Entity has no key metric yet. If you don’t know what I mean by keyMetric, I’m referring to the bug where the keyMetric value users inputted on this page was getting lost. )
- User adds a key metric → AddEditKeyMetricsModal saves by reading the report prop directly → stores that update (entity now has the key metric)
- User fills out a subsequent field (defineInitiative_qualitativeMethod.) → autosave fires → prepareEntityPayload still reads from the original entity (no key metric) → posts the initiative array without the key metric
- The Backend sees these two competing calls and the second call overwrites the first since initiative is an array, thus wiping the key metric

Adding report to the useMemo dependency array ensures that whenever the store is updated after a save, prepareEntityPayload is recalculated and stored in memory again with the fresh report. Subsequent autosaves then correctly spread from the up-to-date entity that includes the keyMetric.

I hope that makes sense! Let me know if theres more detail I can provide or if that was confusing at all :)

ntsummers1 (approved):
Looking solid! I still haven't been able to make initiatives lose their value. That being said we should still keep an eye out just incase. There might be better ways to handle this race condition if it keeps happening

britt-mo (approved):

ajaitasaini (approved):

PR #1447, opened 2026-05-18
Create legacy SAR template from legacy WP data

jessabean:
### Description
This change ensures that even if the feature flag is enabled, if the WP data is from the legacy WP template, the user will see the legacy SAR template when creating a new SAR, rather than the feature flagged version.

❇️ BONUS: also fixes a nasty bug where `null` values for optional entity fields were causing the initiatives page to crash in the new WP/SAR flow. More detail on how to recreate that below.

### Related ticket(s)
CMDCT-5995

### Cloudfront URL
https://d31de7rcce6h74.cloudfront.net/

---
### How to test
**To test this appropriately, you'll need to be able to toggle the feature flag on and off to create reports with "legacy" templates as well as the new template.**

#### Old WP -> Old SAR

1. Run the app locally
2. In [featureFlags.ts](https://github.com/Enterprise-CMCS/macpro-mdct-mfp/blob/main/services/app-api/utils/featureFlags/featureFlags.ts#L5), change line 5 to read:
`JSON.parse({"local": true, "flags": { "abcdReport": false, "wpSarRelease2025": false}})`. Wait for the API to finish building.
3. Create/fill out/submit/approve a Work Plan. It should have the old initiatives flow, with the dashboard and overlays. You can use the `db:seed` utility to do this.
4. Verify quickly that the WP is the legacy version. When clicking "View" on an initiative, you should see the old initiative dashboard view (with Define Initiative, Evaluation Plan, Funding Sources, etc).
5. Create a new SAR using the UI or `db:seed`.
6. View the State or Territory-Specific Initiatives section of the SAR and verify that it is using the old template. Similar to the WP, there should be an initiative dashboard view.

**Old WP**
Screenshot 2026-05-18 at 13-08-11 Money Follows the Person

**Old SAR**
Screenshot 2026-05-18 at 13-11-22 Money Follows the Person

#### New WP -> New SAR
1. In [featureFlags.ts](https://github.com/Enterprise-CMCS/macpro-mdct-mfp/blob/main/services/app-api/utils/featureFlags/featureFlags.ts#L5), change line 5 to read:
`JSON.parse({"local": true, "flags": { "abcdReport": false, "wpSarRelease2025": true}})`. Wait for the API to finish building.
2. Create/fill out/submit/approve a Work Plan. It should have the NEW initiatives flow, without a dashboard, and a single overlay page per initiative.
3. Create a new SAR using the UI or `db:seed`.
4. View the State or Territory-Specific Initiatives section of the SAR and verify that it is using the NEW template. 

#### ❇️ BONUS: recreate the null values bug
If you want to understand why [b393a43](https://github.com/Enterprise-CMCS/macpro-mdct-mfp/pull/1447/commits/b393a43307b1a5c436c117cbec6268152dfcf658) came to be, this is for you. To recreate:

1. Switch to main (not this branch)
2. Make sure the wpSarRelease2025 feature flag is enabled
3. Create an approved WP
4. Then create a **copy over** from that WP (you need this to get the closeout fields in the initiatives page)
5. Edit one of the initiatives. In the Closeout Information section, click into the actual end date field, then leave it (don't enter info). Save and return.
6. Watch the page blow up.

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

### DIFF
--- a/services/app-api/utils/formTemplates/formTemplates.test.ts
+++ b/services/app-api/utils/formTemplates/formTemplates.test.ts
@@ -12,6 +12,7 @@ import {
   getOrCreateFormTemplate,
   isFieldElement,
   isLayoutElement,
+  isLegacySAR,
 } from "./formTemplates";
 // forms
 import { wpReportJson as wp, sarReportJson as sar } from "../../forms";
@@ -328,19 +329,13 @@ describe("Test formTemplateForReportType legacy WP & SAR", () => {
     jest.clearAllMocks();
   });
 
-  const getInitiativesRoute = (formTemplate: ReportJson) =>
-    formTemplate.routes.find(
-      (route) => route.path === "/sar/state-or-territory-specific-initiatives"
-    ) as DynamicModalOverlayReportPageShape | undefined;
-
   it("returns legacy SAR template (with initiatives) when workPlanFieldData is from a legacy WP", async () => {
     const legacyFieldData = { strategy_additionalDetails: "some value" };
     const template = await formTemplateForReportType(
       ReportType.SAR,
       legacyFieldData
     );
-    const initiativesRoute = getInitiativesRoute(template);
-    expect("initiatives" in (initiativesRoute ?? {})).toBe(true);
+    expect(isLegacySAR(template)).toBe(true);
   });
 
   it("returns new SAR template when workPlanFieldData lacks 'strategy_additionalDetails'", async () => {
@@ -350,8 +345,7 @@ describe("Test formTemplateForReportType legacy WP & SAR", () => {
       ReportType.SAR,
       modernFieldData
     );
-    const initiativesRoute = getInitiativesRoute(template);
-    expect("initiatives" in (initiativesRoute ?? {})).toBeFalsy();
+    expect(isLegacySAR(template)).toBeFalsy();
   });
 
   it("does not short-circuit feature flagged templates when workPlanFieldData is absent", async () => {
--- a/services/app-api/utils/formTemplates/formTemplates.test.ts
+++ b/services/app-api/utils/formTemplates/formTemplates.test.ts
@@ -345,7 +345,7 @@ describe("Test formTemplateForReportType legacy WP & SAR", () => {
       ReportType.SAR,
       modernFieldData
     );
-    expect(isLegacySAR(template)).toBeFalsy();
+    expect(isLegacySAR(template)).toBe(false);
   });
 
   it("does not short-circuit feature flagged templates when workPlanFieldData is absent", async () => {
--- a/services/app-api/utils/formTemplates/formTemplates.test.ts
+++ b/services/app-api/utils/formTemplates/formTemplates.test.ts
@@ -323,6 +323,51 @@ describe("Test form contents", () => {
   });
 });
 
+describe("Test formTemplateForReportType legacy WP & SAR", () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  const getInitiativesRoute = (formTemplate: ReportJson) =>
+    formTemplate.routes.find(
+      (route) => route.path === "/sar/state-or-territory-specific-initiatives"
+    ) as DynamicModalOverlayReportPageShape | undefined;
+
+  it("returns legacy SAR template (with entitySteps) when workPlanFieldData is from a legacy WP", async () => {
+    const legacyFieldData = { strategy_additionalDetails: "some value" };
+    const template = await formTemplateForReportType(
+      ReportType.SAR,
+      legacyFieldData
+    );
+    const initiativesRoute = getInitiativesRoute(template);
+    expect("entitySteps" in (initiativesRoute?.initiatives?.[0] ?? {})).toBe(
+      true
+    );
+  });
+
+  it("returns new SAR template when workPlanFieldData lacks 'strategy_additionalDetails'", async () => {
+    (isFeatureFlagEnabled as jest.Mock).mockResolvedValue(true);
+    const modernFieldData = { someOtherKey: "value" };
+    const template = await formTemplateForReportType(
+      ReportType.SAR,
+      modernFieldData
+    );
+    const initiativesRoute = getInitiativesRoute(template);
+    expect(
+      "entitySteps" in (initiativesRoute?.initiatives?.[0] ?? {})
+    ).toBeFalsy();
+  });
+
+  it("does not short-circuit feature flagged templates when workPlanFieldData is absent", async () => {
+    (isFeatureFlagEnabled as jest.Mock).mockResolvedValue(false);
+    await formTemplateForReportType(ReportType.WP);
+    expect(isFeatureFlagEnabled).toHaveBeenCalled();
+
+    await formTemplateForReportType(ReportType.SAR);
+    expect(isFeatureFlagEnabled).toHaveBeenCalled();
+  });
+});
+
 describe("Test compileValidationJsonFromRoutes", () => {
   it("Compiles validation from forms of any kind", () => {
     const result = compileValidationJsonFromRoutes(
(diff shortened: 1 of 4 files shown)


services/ui-src/src/components/tables/getEntityStatus.tsx
```diff
@@ -32,7 +32,8 @@ export const getValidationList = (fields: AnyObject[], entity: AnyObject) => {
   //Loop through the user entered entity to see which nested id the user had selected
   const entityNestedSelection: AnyObject[] = [];
   Object.values(entity).forEach((field: AnyObject) => {
-    if (typeof field === "object") {
+    // typeof null === "object", so guard against null field values
```
jessabean: This is a quick fix to deal with null values (a result of specific [**optional** number fields returning null](https://github.com/Enterprise-CMCS/macpro-mdct-mfp/blob/main/services/ui-src/src/utils/validation/schemas.ts#L98-L99)) that were preventing the entity rows from rendering, thus crashing the Initiatives page. The full story is annoying and sad, and will be addressed in a future task/sprint/spike whatever.

services/app-api/utils/formTemplates/formTemplates.ts
```diff
@@ -51,6 +54,15 @@ export const formTemplateForReportType = async (reportType: ReportType) => {
   const flagsByReportType = flagMap[reportType];
   const flagNames = Object.keys(flagsByReportType);
 
+  // Legacy WP field data should always produce a legacy SAR template, regardless of feature flags
+  const isLegacySAR =
```
bangbay-bluetiger: It's a little confusing to use `isLegacySAR` here and also for the exported function below.

services/app-api/utils/formTemplates/formTemplates.ts
```diff
@@ -143,6 +158,16 @@ export async function getOrCreateFormTemplate(
   }
 }
 
+export const isLegacySAR = (template: ReportJson): boolean => {
```
bangbay-bluetiger, line 161: Where is this being used?

services/ui-src/src/components/tables/getEntityStatus.tsx
```diff
@@ -32,7 +32,8 @@ export const getValidationList = (fields: AnyObject[], entity: AnyObject) => {
...
+    if (field !== null && typeof field === "object") {
```
bangbay-bluetiger: We could use a filter here instead:
```
const entityNestedSelection = Object.values(entity).filter(
  (field) => field !== null && typeof field === "object"
);
```

services/app-api/utils/formTemplates/formTemplates.ts
jessabean: Whoops...as I was iterating on this branch I meant to remove the const above in favor of this one (which is used in the test). Let me clean that up. Thanks for catching!

jessabean: Ah, now I remember. One for creating the right template at report creation time, and one for determining WHICH version of the SAR you are in. I will just rename one to make that more clear.

services/ui-src/src/components/tables/getEntityStatus.tsx
```diff
@@ -41,7 +42,7 @@ export const getValidationList = (fields: AnyObject[], entity: AnyObject) => {
   const entityNestedKeys = entityNestedSelection
     .flat()
     .map((entity) => {
-      return entity.key;
+      return entity?.key;
```
benmartin-coforma: We just made sure these will be truthy. There should be no need for the safe-access operator.
```suggestion
      return entity.key;
```

services/ui-src/src/components/tables/getEntityStatus.test.ts
```diff
@@ -69,6 +69,48 @@ describe("tables/getEntityStatus", () => {
...
+        },
+      ];
+      const entity = {
+        id: "entity1",
+        type: EntityType.INITIATIVE,
+        field1: null,
+        field2: [null, { key: "choice1key" }],
+      };
+
+      expect(() => getValidationList(fields, entity)).not.toThrow();
```
benmartin-coforma: This expectation is redundant with the line below; we already expect it not to throw, in that we're calling it.

services/ui-src/src/components/tables/getEntityStatus.tsx
jessabean: Yikes, turns out we do need this after all

services/app-api/utils/formTemplates/formTemplates.test.ts
```diff
@@ -323,6 +324,40 @@ describe("Test form contents", () => {
...
+  });
+
+  it("returns new SAR template when workPlanFieldData lacks 'strategy_additionalDetails'", async () => {
+    (isFeatureFlagEnabled as jest.Mock).mockResolvedValue(true);
+    const modernFieldData = { someOtherKey: "value" };
+    const template = await formTemplateForReportType(
+      ReportType.SAR,
+      modernFieldData
+    );
+    expect(isLegacySAR(template)).toBeFalsy();
```
bangbay-bluetiger: Why use `toBeFalsy()` instead of `toBe(false)`?

services/app-api/utils/formTemplates/formTemplates.ts
```diff
@@ -143,6 +158,16 @@ export async function getOrCreateFormTemplate(
...
+  // If you duplicate this function in ui-src, update it there too!
```
bangbay-bluetiger, line 162: Are there plans to use this function in `ui-src`?

services/ui-src/src/components/tables/getEntityStatus.tsx
```diff
@@ -30,12 +30,10 @@ export const getValidationList = (fields: AnyObject[], entity: AnyObject) => {
...
   //Loop through the user entered entity to see which nested id the user had selected
-  const entityNestedSelection: AnyObject[] = [];
-  Object.values(entity).forEach((field: AnyObject) => {
-    if (typeof field === "object") {
-      entityNestedSelection.push(field);
-    }
-  });
+  // typeof null === "object", so we need to guard against null field values
+  const entityNestedSelection = Object.values(entity).filter(
+    (field) => field !== null && typeof field === "object"
```
bangbay-bluetiger, line 35: In your test, `field2: [null, { key: "choice1key" }]` passes the `null` through because we would need to change this function to recursively filter through arrays.

services/ui-src/src/components/tables/getEntityStatus.tsx
benmartin-coforma: I don't understand how that's possible.

services/app-api/utils/formTemplates/formTemplates.test.ts
jessabean: Because I wasn't looking close enough... 😆 Updating!

services/ui-src/src/components/tables/getEntityStatus.tsx
jessabean: The mock in the test doesn't actually represent our implementation of stored form data, which took me a while to sit down and console.log. So I've updated the mock!

services/app-api/utils/formTemplates/formTemplates.ts
jessabean: I could see a future where that could be the case (the need to do a check to determine if the template is a legacy SAR template in the UI), but at the moment, I'm not sure we have a place to store things that are needed both in `app-api` and `ui-src`. The pattern is to maintain both separately. So this comment was kind of a heads up.

services/ui-src/src/components/tables/getEntityStatus.tsx
jessabean: nm, was a failing test but the mock was incorrect. 😓

bangbay-bluetiger (approved):
Tested, and works as expected. 🎉

PR #1452, opened 2026-05-21
Add error message for EntityModalTable

bangbay-bluetiger:
### Description

- Use `InlineError` component in `EntityModalTable`
- Modify `Form` component submission to check for table errors
- Create a `isTableField` helper and replace wherever `f.forTableOnly` was used with it

### Related ticket(s)

CMDCT-6007, CMDCT-5333

---
### How to test

1. https://d3mo5qq8jnsra1.cloudfront.net/
2. Create a Work Plan
3. Start an initiative, fill out every field except Key Metrics
4. Submit the form, confirm it shows an error for Key Metrics
    Screenshot 2026-05-21 at 4 25 20 PM
5. Go back to Initiatives dashboard, confirm initiative has incomplete status
    Screenshot 2026-05-21 at 4 28 30 PM
6. Add a Key Metric
7. Submit the form, confirm initiative has complete status
    Screenshot 2026-05-21 at 4 29 16 PM

---
### Pre-review checklist

---
#### Review


services/ui-src/src/components/tables/EntityModalTable.tsx
```diff
@@ -65,7 +96,8 @@ export const EntityModalTable = ({
     setCurrentEntityId(entityId);
     keyMetricsModalOnOpenHandler();
   };
-  const closeModal = () => {
+  const closeModal = (closedAfterSave?: boolean | BaseSyntheticEvent) => {
+    if (closedAfterSave === true) setErrorMessage(undefined);
```
britt-mo, line 101: I think this is the first time I've seen `BaseSyntheticEvent`, could you explain why it's necessary here?

services/ui-src/src/components/tables/ErrorMessage.tsx
```diff
@@ -0,0 +1,42 @@
...
+// assets
+import unfinishedIcon from "assets/icons/icon_error_circle.png";
+
+export const ErrorMessage = ({ id, message }: Props) => {
+  if (!message) return null;
+
+  return (
+    <Flex
+      aria-atomic="true"
+      aria-live="assertive"
```
britt-mo: I think it's the first time I'm seeing these two attributes as well, are these ones that we will/should start using more often? Just curious.

services/ui-src/src/components/modals/AddEditKeyMetricsModal.tsx
```diff
@@ -67,7 +67,7 @@ export const AddEditKeyMetricsModal = ({
 
   const handleSubmit = async (enteredData: AnyObject) => {
     if (viewOnly) {
-      modalDisclosure.onClose();
+      modalDisclosure.onClose(false);
```
britt-mo, line 70: why did these modal disclosures need the explicit `true` and `false`?

services/ui-src/src/components/tables/ErrorMessage.tsx
```diff
@@ -0,0 +1,42 @@
+import { ReactNode } from "react";
+// components
+import { Flex, Image, Text } from "@chakra-ui/react";
+// assets
+import unfinishedIcon from "assets/icons/icon_error_circle.png";
+
+export const ErrorMessage = ({ id, message }: Props) => {
+  if (!message) return null;
```
jessabean: If what I understand about aria-live regions is correct, then the aria attributes as written will be unreliable, because the Flex component itself is dynamic (appears/disappears from the DOM based on the truthiness of the message prop). So I think we would need the Flex component always present in the DOM and have another wrapper element around the contents inside, that mounts/unmounts as needed. E.g., 

```
const messageBody = () => {
  if (!message) return null;

  return (
    <>
      // Img, Text, message etc go here
    </>
  )
}

return (
  <Flex role="region" aria-atomic="true" aria-live="assertive"....> // this always appears in DOM
    { messageBody}
  </Flex>
```

karla-vm (dismissed):
LGTM, tested changes locally and updated content ✅ not a blocker, but definitely curious to learn more re: @britt-mo's questions about new attributes.

services/ui-src/src/components/tables/EntityModalTable.tsx
bangbay-bluetiger: We call `modalDisclosure.onClose` in three places per modal:

1. With the modal X close button
2. After successful onSubmit
3. On return (when submitting is disabled)

For 2 and 3, I've opted to return `true` or `false` from `modalDisclosure.onClose` (I'll explain why in a different comment), but by default it returns an event. `BaseSyntheticEvent` was suggested from "Infer parameter types from usage." I'm adding the type here to explain whey I'm using `if (closedAfterSave === true)` and not `if (closedAfterSave)`.

services/ui-src/src/components/tables/ErrorMessage.tsx
bangbay-bluetiger: I modeled this component after the error message that CMSDS produces, which has these attributes. I did not look into if I could reuse the CMSDS error component.

CMSDS:
Screenshot 2026-05-27 at 7 56 32 AM

This `ErrorMessage` component:
Screenshot 2026-05-27 at 7 56 47 AM

services/ui-src/src/components/modals/AddEditKeyMetricsModal.tsx
bangbay-bluetiger: This was a bit of hack as I needed a way for `EntityModalTable` to clear an error message in this scenario:

- User hits submit button when there isn't a Key Metric, error appears in the `EntityModalTable` table
- User adds a Key Metric
- `EntityModalTable` needs to know that if validation has passed to clear the error message

I thought about having the component listen to `report.fieldData` via useEffect, but I went with `modalDisclosure.onClose(true)` instead, since having a successful Key Metric save is enough to convey that the validation passed.

services/ui-src/src/components/tables/ErrorMessage.tsx
bangbay-bluetiger: Okay, I see that. I can address that in this PR or would making the change in another PR be okay?

services/ui-src/src/components/tables/ErrorMessage.tsx
bangbay-bluetiger: I can reuse `InlineError` instead of introducing this new component. I'll go ahead an swap them.

services/ui-src/src/components/tables/ErrorMessage.tsx
bangbay-bluetiger: I removed this component and used CMSDS `InlineError` component. The wrapper component should be present when there is no error message, like the other form fields.

karla-vm (approved):

britt-mo (approved):
Tested locally and worked as expected!

Also, I really appreciate the detailed responses to my questions. Thanks!

PR #1468, opened 2026-06-04
Cmdct 5785 - updates to dashboard

britt-mo:
### Description
per the card:

Farthest right column header should have the text `Actions`
The icon on the far left to edit a report name should:
Change from an icon to the text `Edit reporting` reporting visually, and with an accessible name (accName) of `Edit reporting of <report>`
Move this link from its own column, to being the first link in the `Actions` column

Also: 
Admin users will no longer see the edit modal for the program creation. [Card here](https://jiraent.cms.gov/browse/CMDCT-6199)

[figma here](https://www.figma.com/design/cNFWWAgpTPfir1kglCnOJ2/MFP-FRF--Internal-?node-id=1-2&p=f&t=daRkVDFU3RLSJZgN-0)

### Related ticket(s)

CMDCT-5785

---
### How to test

This is easier to test locally so you can run the` yarn db:seed` command and generate a workplan and a sar.
Confirm that WP doesn't have changes. 
Confirm that the SAR and Financial Report dashboards look like the figma. 
Login as an admin and confirm the changes there too.

### Notes

---
### Pre-review checklist

---
#### Review

#### Security


britt-mo:
When a report has both `Unlock` and `Archive`, the columns get compressed (with the new view/edit reporting button) so the row gets taller because of the long report name, width limit of the columns, and expansion of the `Actions` column. At least that's how I understand it. Not sure what the fix should be here.

Screenshot 2026-06-04 at 10 29 06 AM

services/ui-src/src/components/pages/Dashboard/DashboardTable.test.tsx
```diff
@@ -1,7 +1,70 @@
...
+
+    test("should show View Reporting when user is admin", () => {
+      render(
+        <RouterWrappedComponent>
+          <DashboardTable {...createProps({ isAdmin: true })} />
+        </RouterWrappedComponent>
+      );
+
+      expect(
+        screen.getByRole("button", { name: /view reporting/i })
```
bangbay-bluetiger: I think we should do exact matches in these tests. But, it looks like we have a bug in that this is labeled `View Reporting of [object Object]` because we're passing in the `report` to the aria-label rather than a string.

https://github.com/Enterprise-CMCS/macpro-mdct-mfp/blob/58b520f9b9c16b3a6de3e294dcfc8bd67acb301a/services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx#L231

services/ui-src/src/components/pages/Dashboard/DashboardTable.test.tsx
```diff
@@ -68,4 +131,117 @@ describe("<DashboardTable />", () => {
...
+            report={mockWPFullReport}
+            reportId="test-id"
+            isStateLevelUser={true}
+            entering={false}
+            enterSelectedReport={() => {}}
+          />
+        </RouterWrappedComponent>
+      );
+      const button = container.querySelector("button");
+      expect(button).toHaveStyle({ width: "auto" });
```
bangbay-bluetiger: Could skip using `container`, and just use `screen`:
```
expect(screen.getByRole("button")).toHaveStyle({ width: "auto" });
```
Comment also applies to other places `container` is used.

services/ui-src/src/components/pages/Dashboard/DashboardTable.test.tsx
```diff
@@ -68,4 +131,117 @@ describe("<DashboardTable />", () => {
...
+          <ActionButton
+            report={mockSARFullReport}
+            reportId="test-id"
+            isStateLevelUser={true}
+            entering={false}
+            enterSelectedReport={() => {}}
+          />
+        </RouterWrappedComponent>
+      );
+      expect(screen.getByRole("button", { name: /Edit/i })).toBeVisible();
```
bangbay-bluetiger: Use exact match instead of regex match:
```
expect(screen.getByRole("button", { name: "Edit 2024 Period 1 report" })).toBeVisible();
```
Comment also applies to other tests using regex match.

services/ui-src/src/components/tables/Table.test.tsx
```diff
@@ -23,5 +25,51 @@ describe("<Table />", () => {
...
+        headRow: ["Header 1", { title: "Actions", colSpan: 2 }],
+        bodyRows: [],
+      };
+      const { container } = render(
+        <RouterWrappedComponent>
+          <Table content={contentWithColSpan} />
+        </RouterWrappedComponent>
+      );
+      const headerCells = container.querySelectorAll("th");
+      expect(headerCells[1]).toHaveAttribute("colspan", "2");
```
bangbay-bluetiger: You can use `screen` here:
```
expect(screen.getByRole("columnheader", { name: "Actions" })).toHaveAttribute("colspan", "2");
``` 
Generally it's better to use the testing library queries instead of DOM queries because it's built-in a11y testing.

services/ui-src/src/components/tables/Table.test.tsx
```diff
@@ -23,5 +25,51 @@ describe("<Table />", () => {
...
+          { title: "Object header", colSpan: 2, align: "center" as const },
+        ],
+        bodyRows: [],
+      };
+      const { container } = render(
+        <RouterWrappedComponent>
+          <Table content={mixedContent} />
+        </RouterWrappedComponent>
+      );
+      const headerCells = container.querySelectorAll("th");
```
bangbay-bluetiger: You can use:
```
const headerCells = screen.getAllByRole("columnheader");
```

services/ui-src/src/forms/adminDashSelector/adminDashSelector.ts
```diff
@@ -20,7 +20,7 @@ const reportChoices = [
     label: "MFP Semi-Annual Progress Report (SAR)",
   },
   {
-    id: "Financial Reporting",
+    id: "FINANCIAL-REPORT",
```
bangbay-bluetiger: Should this match our report types? `FINANCIAL_REPORT`

Then, related, should we use `ReportType.FINANCIAL_REPORT` here?

services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
```diff
@@ -28,102 +26,122 @@ export const DashboardTable = ({
...
+          {/* Admin: Submission count */}
+          {isAdmin && (
+            <Td data-testid="dashboard-submission-count">
+              {!report.submissionCount || report.submissionCount === 0
+                ? 1
+                : report.submissionCount}
+            </Td>
+          )}
+          {/* Edit Button */}
+          {isNonWorkPlan && (
```
bangbay-bluetiger: Instead of having these actions in separate `<Td>`, it would give us more flexibility to put everything in a single `<Td>` and use flex CSS to space out the buttons. I would check with design, but I think having the buttons stacked 2x2 would look okay.

```
<Td>
  <Flex>
    {/* Edit Button */}
    {isNonWorkPlan && (
      <EditReportButton // This is wrapped in a <Td>, that'll need to be removed/changed to Flex
        report={report}
        openCreateReportModal={openCreateReportModal}
        isAdmin={isAdmin}
        sxOverride={sxOverride}
      />
    )}
    {/* Action Buttons */}
    <Flex sx={actionCellSx}>
      <ActionButton
        report={report}
        reportId={reportId}
        isStateLevelUser={isStateLevelUser}
        entering={entering}
        enterSelectedReport={enterSelectedReport}
      />
    </Flex>
    {isAdmin && (
      <>
        {
          <AdminReleaseButton // This is wrapped in a <Td>, that'll need to be removed/changed to Flex
            report={report}
            reportType={reportType}
            reportId={reportId}
            releaseReport={releaseReport}
            releasing={releasing}
            sxOverride={sxOverride}
          />
        }
        {isArchivable(reportType) && !report?.associatedSar && (
          <AdminArchiveButton // This is wrapped in a <Td>, that'll need to be removed/changed to Flex
            report={report}
            reportType={reportType}
            reportId={reportId}
            archive={archive}
            releaseReport={releaseReport}
            releasing={releasing}
            sxOverride={sxOverride}
          />
        )}
      </>
    )}
  </Flex>
</Td>
```

cy-by (approved):
Reviewed locally with both state and admin users. Things look good :shipit:

bangbay-bluetiger (approved):

PR #1476, opened 2026-06-10
CMDCT-6152 508 issue - error messages not announced

cy-by:
### Description

Because we use a HTML `<select>` for our dropdown, we need to manually handle connecting error messages to it, style it with a red border, and manually change `aria-invalid`.

### Related ticket(s)

CMDCT-6152

---
### How to test
ℹ️ If you're having a hard time dealing with VoiceOver, reach out and I can help!

- Run this PR locally or use the [deployed env](https://doskmvjyae2sa.cloudfront.net/)
- Navigate to FRF dashboard (at `/financial-report`)
- Click the `Start MFP Financial Report` button
- With the `Start MFP Financial Report` modal open, [turn on VoiceOver](https://support.apple.com/guide/voiceover/turn-voiceover-on-or-off-vo2682/mac) (if you're on macOS)
- Use the `Tab` key to navigate to the dropdown fields in the modal
- Once you blur one of the required fields, return to the errored field, you should be able to hear that the error message is read
- For the two required dropdown fields in the modal, you should be able to hear the error message when the error message is visible and you have that field focused

### DIFF
--- a/services/ui-src/src/components/fields/DropdownField.tsx
+++ b/services/ui-src/src/components/fields/DropdownField.tsx
@@ -106,8 +106,9 @@ export const DropdownField = ({
   const formErrorState: AnyObject = form?.formState?.errors;
   const errorMessage = formErrorState?.[name]?.value?.message as ReactNode;
   const parsedHint = hint ? parseCustomHtml(hint) : undefined;
-  const ariaDescribedBy = parsedHint ? `${name}-hint` : undefined;
+  const ariaDescribedBy = `${name}__error${parsedHint ? ` ${name}-hint` : ""}`;
   const nestedChildClasses = nested ? "nested ds-c-choice__checkedChild" : "";
+  const selectClasses = `ds-c-field${errorMessage ? " ds-c-field--error" : ""}`;
   const labelClass = !label ? "no-label" : "";
   const labelText =
     label && styleAsOptional ? labelTextWithOptional(label) : label;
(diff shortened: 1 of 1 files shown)


services/ui-src/src/components/fields/DropdownField.tsx
```diff
@@ -106,8 +106,9 @@ export const DropdownField = ({
   const formErrorState: AnyObject = form?.formState?.errors;
   const errorMessage = formErrorState?.[name]?.value?.message as ReactNode;
   const parsedHint = hint ? parseCustomHtml(hint) : undefined;
-  const ariaDescribedBy = parsedHint ? `${name}-hint` : undefined;
+  const ariaDescribedBy = `${name}__error${parsedHint ? ` ${name}-hint` : ""}`;
   const nestedChildClasses = nested ? "nested ds-c-choice__checkedChild" : "";
+  const selectClasses = `ds-c-field${errorMessage ? " ds-c-field--error" : ""}`;
```
cy-by, line 111: For ternaries that have a string that should be there in both cases, I have it listed first, then use the ternary for the dynamic string. Open to a different approach/syntax.

E.g. for `ariaDescribedBy`, `${name}__error` will always be used, then `${name}-hint` is added after if it exists.

services/ui-src/src/components/fields/DropdownField.tsx
```diff
@@ -120,17 +121,17 @@ export const DropdownField = ({
...
       {parsedHint && <Hint id={`${name}-hint`}>{parsedHint}</Hint>}
-      {errorMessage && <InlineErrorShim>{errorMessage}</InlineErrorShim>}
+      <InlineErrorShim id={`${name}__error`}>{errorMessage}</InlineErrorShim>
       <select
         name={name}
         id={name}
         aria-describedby={ariaDescribedBy}
         aria-label={ariaLabel}
-        aria-invalid="false"
+        aria-invalid={!!errorMessage}
```
cy-by, line 130: Since this is a custom `<select>` and not a component from CMS DS, we need to change this attribute ourselves.

bangbay-bluetiger (approved):
Nice!

karla-vm (approved):

PR #1479, opened 2026-06-11
Add multiline option to DynamicField

bangbay-bluetiger:
### Description

- Add `multiline` and `rows` prop to `DynamicField`
- `rows` will only be recognized if `multiline` is `true`
- Add `multiline: true` to Initiatives route

### Related ticket(s)

CMDCT-6176

---
### How to test

1. https://dcx8qwkqqeefv.cloudfront.net/
2. Create a Work Plan
3. Create an Initiative
4. Under Key Activities, the dynamic fields should be textareas

Screenshot 2026-06-12 at 1 08 40 AM
Screenshot 2026-06-12 at 1 08 53 AM

---
### Pre-review checklist

---
#### Review

### DIFF
--- a/services/ui-src/src/components/fields/DynamicField.tsx
+++ b/services/ui-src/src/components/fields/DynamicField.tsx
@@ -228,30 +228,42 @@ export const DynamicField = ({
       )}
 
       {displayValues.map((field: DynamicFieldShape, index: number) => {
+        const errorMessage = fieldErrorState?.[index]?.name?.message;
+        const hasError = Boolean(errorMessage);
+        const textareaStyle = hasError
+          ? sx.removeBoxTextareaError
+          : sx.removeBoxTextarea;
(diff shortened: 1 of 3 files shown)


services/ui-src/src/components/fields/DynamicField.tsx
```diff
@@ -335,4 +363,10 @@ const sx = {
       width: "100%",
     },
   },
+  dynamicFieldInput: {
+    alignItems: "flex-end",
+  },
+  dynamicFieldTextarea: {
+    alignItems: "center",
```
jessabean, line 370: Non-blocking, just venting: There is much redundancy in this component's CSS that I think could be fixed with style/markup refactor using Grid instead of Flex (so we could reduce the styling that is textarea vs input specific). But le sigh, that's a problem for another day...

jessabean (approved):
It does the thing! :shipit:

ajaitasaini (approved):

karla-vm (approved):

PR #1506, opened 2026-07-06
cmdct-6241 - display key metrics table header only if a key metric exists

britt-mo:
### Description

If there any no Key Metrics, then the headers should not display, like so:

Screenshot 2026-07-06 at 3 52 24 PM

### Related ticket(s)

CMDCT-6214

---
### How to test
link: https://d1sntngvgiogch.cloudfront.net/
Create a WP, create an initiative, see that the headers don't display until you've added a Key Metric. Deleting the Key Metric should also revert to hiding the headers.

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

### DIFF
--- a/services/ui-src/src/components/tables/EntityModalTable.tsx
+++ b/services/ui-src/src/components/tables/EntityModalTable.tsx
@@ -141,8 +141,8 @@ export const EntityModalTable = ({
   };
   const dynamicRowEntries = getDynamicRowEntries();
 
-  // Hide headers for a dynamic-rows table until at least one row is added
-  const showHeadRows =
+  // Hide the table for a dynamic-rows table until at least one row is added
+  const showTable =
     !dynamicRowsTemplate || bodyRows.length > 0 || dynamicRowEntries.length > 0;
 
   const sharedCellProps = {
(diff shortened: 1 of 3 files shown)


services/ui-src/src/components/tables/EntityModalTable.tsx
```diff
@@ -170,16 +188,17 @@ export const EntityModalTable = ({
...
-            generateRows({
-              row,
-              rowIndex,
-              section: "thead",
-              showEditHeader: true,
-              styleAsOptionalHeadRows: styleAsOptionalHeadRows,
-              ...sharedCellProps,
-            })
-          )}
+          {showHeadRows &&
```
bangbay-bluetiger: I think we should just not render the table at all if there are no body rows, something like:
```
const showTable = !dynamicRowsTemplate || bodyRows.length > 0 || dynamicRowEntries.length > 0;

{showTable && <Table />}
{!showTable && <Text>{verbiage?.emptyTableMessage}</Text>}
```

bangbay-bluetiger (approved):

ajaitasaini (approved):
just tested, lgtm!

karla-vm (approved):

PR #1512, opened 2026-07-13
[CMDCT-6205] Add feature flag leaks actions

bangbay-bluetiger:
### Description

This PR adds two GitHub actions to find code referencing feature flag names that do not exist in Production. In `ui-src`, it looks for flags called with `useFlags`. In `app-api`, it looks for flags called with `isFeatureFlagEnabled`.

#### `pr-feature-flag-leaks`

- This action runs on every PR create/update
- If leaks are found, it will post a comment to the PR
    - If none are found, it won't post a new comment
    - If there is an existing comment and the leaks are fixed, the comment will update with a "No feature flag leaks found" message
- If leaks are found, it will not block the current PR
- The leaks comment is to raise awareness and is not necessarily caused by changes in the current PR

Screenshot 2026-07-14 at 12 16 17 PM

Screenshot 2026-07-14 at 2 09 17 PM

#### `feature-flag-leaks`

- This action will run every Monday morning, but can also be manually run
- This action is intended to capture leaks if flags are deleted in LaunchDarkly, but there are no active PRs to catch the change
- If leaks are found, it will post a message to the [#macpro-mdct-mfp-alerts](https://cmsgov.enterprise.slack.com/archives/C06R1622XBQ) channel
- If none are found, it won't post anything

Screenshot 2026-07-14 at 12 16 25 PM

### Related ticket(s)

CMDCT-6205

---
### How to test

1. Trigger the PR comment by adding non-existent flags and pushing to this branch: https://github.com/Enterprise-CMCS/macpro-mdct-mfp/pull/1512/changes/87b74cfe33cd08463d067a51cec2155e719e8961
2. Trigger the Slack message by adding a push trigger for this branch (otherwise the action can't be manually triggered until it's merged into `main`): https://github.com/Enterprise-CMCS/macpro-mdct-mfp/pull/1512/changes/8295414e8118d68db2c32fa102a5522895a49560
3. Verify PR comment and Slack message were posted
4. Remove the non-existent flags and push to this branch
5. Verify PR comment updates to `No feature flag leaks found` message
6. No new Slack message should be posted
7. Remove the push trigger and push to this branch

---
### Pre-review checklist

---
#### Review

### DIFF
--- a/.github/feature-flags.ts
+++ b/.github/feature-flags.ts
@@ -110,7 +110,8 @@ export function formatPrComment(featureFlagNames: string[]) {
       commentTag,
       "### Feature Flag Leaks",
       "",
-      `⚠️ ${warningMessage}\n`,
+      `⚠️ ${warningMessage}`,
+      "",
       "",
       "|File|Line|Flag|",
       "|----|----|----|",
@@ -134,7 +135,7 @@ export function formatSlackMessage(featureFlagNames: string[]) {
       }
     );
 
-    return [`:warning: ${warningMessage}\n`, ...formattedLines].join("\n");
+    return [`:warning: ${warningMessage}`, "", ...formattedLines].join("\n");
   }
 
   return "";
--- a/.github/feature-flags.ts
+++ b/.github/feature-flags.ts
@@ -7,18 +7,25 @@ function getLinesWithFlags(featureFlagNames: string[]) {
   const repoName = process.env.GITHUB_REPOSITORY;
   const repoBaseUrl = `${serverUrl}/${repoName}/blob/${commitSha}/`;
 
+  /**
+   * Regex matches for:
+   * useFlags()?.someFlagName
+   * useFlags().someFlagName
+   * isFeatureFlagEnabled("someFlagName")
+   */
   const searchPatterns = [
-    /useFlags\(\)(?:\?\.)?([a-zA-Z0-9_]+)/g,
-    /isFeatureFlagEnabled\(['"`]([a-zA-Z0-9_]+)['"`]\)/g,
+    /useFlags\(\)(?:\?)?\.([a-zA-Z0-9_-]+)/g,
+    /isFeatureFlagEnabled\(["]([a-zA-Z0-9_-]+)["]\)/g,
   ];
 
+  // Grep services folders for matches, excluding test files
   const rawGrepOutput = execFileSync(
     "git",
     [
       "grep",
       "-n",
       "-E",
-      "useFlags\\(\\)|isFeatureFlagEnabled\\((['\"`])[^'\"`]+(['\"`])\\)",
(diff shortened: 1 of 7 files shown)


karla-vm (dismissed):

.github/feature-flags.ts
```diff
@@ -0,0 +1,144 @@
...
+      }
+    );
+
+    return [
+      ":warning: This app references feature flags that do not exist in Production. Verify whether code is valid or if flag needs to be created.\n",
+      ...formattedLines,
+    ].join("\n");
+  }
+
+  return "";
```
jessabean, line 147: Question: This returns empty string for the no leaks condition, while the PR comment function returns `null`. Just curious as to why they are different?

amanda-forrester (commented):
I tested both the bot PR comments and the slack notification and they both worked as expected (I created a draft PR to test). I will review the code next!

.github/feature-flags.ts
```diff
@@ -0,0 +1,170 @@
...
+          const flagName = match[1];
+          if (!featureFlagNames.includes(flagName)) {
+            lines.push({ fileName, flagName, lineNumber, repoBaseUrl });
+          }
+        }
+      });
+    });
+
+    return lines;
+  } catch {
```
amanda-forrester: It's possible I've missed some error handling down the line, it seems here like it won't distinguish between between `git grep` finding no matches (expected) and any other failure. If that's the case, do you think it makes sense to be more explicit?

.github/feature-flags.ts
bangbay-bluetiger: I made this return an empty string because it's used to determine whether to post the Slack message within the workflow YML:
https://github.com/Enterprise-CMCS/macpro-mdct-mfp/blob/0660f4fbc0c030509fb4a78f06e792c6f51460ca/.github/workflows/feature-flag-leaks.yml#L36-L41

A null might work here the same, but I didn't try it out.

For the other action, whether to post PR comment is determined in the TS and returning null seemed to be clearest:

https://github.com/Enterprise-CMCS/macpro-mdct-mfp/blob/0660f4fbc0c030509fb4a78f06e792c6f51460ca/.github/pr-feature-flag-leaks.ts#L46-L53

.github/feature-flags.ts
bangbay-bluetiger: This try doesn't distinguish between errors vs no matches. I could put a console error here so it prints out to the GitHub action log and we can examine it if needed. But, I'm not sure about needing to handle the outcomes differently.

.github/feature-flags.ts
bangbay-bluetiger: I changed the return to null and it printed it out. I'm going to leave it as an empty string return.

Screenshot 2026-07-20 at 12 10 02 PM

ajaitasaini (approved):
tested this out last week, lgtm!

amanda-forrester (approved):

PR #1519, opened 2026-07-22
Cmdct 6280 - add required yes/no question to close-out section, buggy save fix

britt-mo:
### Description
The PR adds a required question to the close-out section, asking the user if they want to close out the initiative:

Screenshot 2026-07-22 at 3 13 03 PM

When you select `Yes`, the optional close-out questions will display:

Screenshot 2026-07-22 at 3 13 16 PM

Fix: Initiatives in the SAR, created from a copied over WP should look like this, where the user can enter a closed initiative and fill it out on the next screen:
Screenshot 2026-07-22 at 3 17 24 PM

Screenshot 2026-07-22 at 3 18 52 PM

Another fix in this PR: fixes bugginess/race conditions around manual saving and autosaving.

**AI summary per commit:**

two `wip` commits reworked the initiative close-out feature:

Close-out becomes a Yes/No radio — initiatives.ts: the three flat close-out fields (projected/actual end date, initiative status) are now nested children under a new `closeOutInformation_closeOut radio` ("Do you want to close out this initiative?"), with `parentOptionId` set on nested validations.
Nested field types — routes.ts: extracted NestedFormField/NestedFormFieldChoice/NestedFormFieldProps, and `NestedFormFieldProps.choices` enables arbitrarily deep nesting.
PDF export — `ExportedModalOverlayReportSection.tsx`: recursive flattenFields so deeply nested close-out fields appear in the PDF.
`isClosedInitiative`— initiatives.ts: recognizes the new radio ("Yes"), with the old fields as fallback.
Form CSS — Form.tsx: choice-label alignment rule.

`8d3e2814` — autosave serialization. Fixes the last-write-wins race where a field's blur autosave and the "Save & return" submit both PUT the full entity array. Adds a serial write queue `enqueueWrite` in autosave.ts; the 5 field components enqueue saves; EntityProvider reads the freshest report and returns a new array; the two submit handlers `await waitForAutosavesAndGetReport` ; 7 test mocks updated.

`e3716144` — completion status. getEntityStatus.tsx: skips nested children with optional validation, and skips `forCopyoverOnly`fields for non-copied entities, so WP initiatives aren't marked incomplete over SAR-only close-out fields. With tests.

`25861a9c` — SAR editable. EntityDetailsOverlayV2.tsx: gates the closed-initiative read-only lock on Work Plan `isInitiativeClosed && isWP` and keeps autosave on in the SAR, so state users can still fill out closed-out initiatives (which stay showing incomplete). With tests.

### Related ticket(s)

CMDCT-6280

---
### How to test

Running locally is the easiest way. 
Run `yarn db:seed` and create an approved WP and a SAR report.
Go in locally and manually add a copied over report. 
Test that there is a new required yes/no question in the close-out section. Test functionality of the initiatives in that report. 
Submit it, approve it, and then create a new SAR. Confirm you can edit previously closed out initiatives

### Notes

---
### Pre-review checklist

---
#### Review

#### Security


services/ui-src/src/components/export/ExportedModalOverlayReportSection.tsx
```diff
@@ -415,19 +415,28 @@ const EntityFieldsTable = ({
...
-    const choices = field.props?.choices;
-    if (Array.isArray(choices)) {
-      choices.forEach((choice: any) => {
-        if (Array.isArray(choice.children)) {
-          flattenedFields.push(...choice.children);
-        }
-      });
-    }
-  });
+  // Flatten nested fields from choices.children into main fields array.
```
bangbay-bluetiger, line 418: Not necessarily related to this PR but this component is marked deprecated because it should only be used for the v1 version of the initiatives. So, I was a little confused why there were changes in this file for this PR.

The initiative v2 code would be better extracted to a separate component and called at the `ExportedReportWrapper` level, similar to how it's done in `ReportPageWrapper`:

https://github.com/Enterprise-CMCS/macpro-mdct-mfp/blob/5d65f3614fafbfe28851e9a61579a7140286139b/services/ui-src/src/components/reports/ReportPageWrapper.tsx#L48-L63

services/ui-src/src/components/export/ExportedModalOverlayReportSection.tsx
```diff
@@ -415,19 +415,28 @@ const EntityFieldsTable = ({
...
+  // Recurse so that fields nested more than one level deep (e.g. the close-out
+  // termination reason / alternate funding text fields, which live under a
+  // choice of closeOutInformation_initiativeStatus) are also included.
+  const flattenFields = (
+    fieldList: (FormField | FormLayoutElement)[]
+  ): (FormField | FormLayoutElement)[] => {
+    const result: (FormField | FormLayoutElement)[] = [];
+    fieldList.forEach((field) => {
```
bangbay-bluetiger: Can you use `flatMap` here?

services/ui-src/src/utils/reports/initiatives.ts
```diff
@@ -4,7 +4,10 @@ import { AnyObject, FormJson } from "types";
 import { isFieldElement } from "utils";
 
 export const isClosedInitiative = (data?: AnyObject) => {
+  const closeOutSelected =
+    data?.closeOutInformation_closeOut?.[0]?.value === "Yes";
   return Boolean(
+    closeOutSelected ||
```
bangbay-bluetiger: A `Yes` answer seems sufficient to check for if the initiative is closed. We can remove checking for `closeOutInformation_actualEndDate` or `closeOutInformation_initiativeStatus`.

ntsummers1:
Additional context on these changes: 

The work in this PR is for adding a close out section for initiatives in the work plan that were copied over from the prior initiative:

- It adds a question at the bottom of each initiative that was copied and asks if you to “Close Out” the initiative if its no longer relevant. 
  - If you do choose to close it out, there are 3 optional questions a user can fill out as to why. 

- When you submit the initiative as closed out, it will then show up on the initiatives dashboard as such. 
- Finally, validation on the review and submit page should require that if a initiative was of the 3 required ones, a new one needs to opened and submitted before we mark that section as complete


We also noticed a bug in that if you click ‘Yes’ to Close Out, and then immediately click save and return, your changes are immediately in a race. The choice list fires and the button fires at the same time. That wouldn’t normally be a problem but in MFP we add data upon clicking Save and Return. Sooo those changes get potentially lost in the race condition and we see weirdness. We fixed this by making writes to the db essentially queue. If a change is in flight, it’ll hold its change until that prior change is done. Based on my testing everything seems fine with it, but it is a slightly change. I’d love your feedback here. Was this too far? Is there another way to go about this (Like turning off autosave on those overlay pages?) Would love any and all thoughts there!

services/ui-src/src/components/overlays/EntityDetailsOverlayV2.test.tsx
```diff
@@ -66,6 +67,11 @@ const entityDetailsOverlayComponent = (
 describe("<EntityDetailsOverlayV2 />", () => {
   beforeEach(() => {
     jest.clearAllMocks();
+    mockedUseStore.mockReturnValue({
+      ...mockStateUserStore,
+      ...mockEntityStore,
+      setAutosaveState: jest.fn(),
```
bangbay-bluetiger: `setAutosaveState` isn't being used, so don't need to add this to the mock.

services/ui-src/src/components/overlays/EntityDetailsOverlayV2.test.tsx
```diff
@@ -177,6 +183,46 @@ describe("<EntityDetailsOverlayV2 />", () => {
     expect(button).toBeVisible();
   });
 
+  test("keeps a closed initiative editable in the SAR", async () => {
+    mockedUseStore.mockReturnValue({
+      ...mockStateUserStore,
+      ...mockEntityStore,
+      report: { reportType: ReportType.SAR } as any,
+      setAutosaveState: jest.fn(),
```
bangbay-bluetiger: `setAutosaveState` isn't being used, so don't need to add this to the mock.

services/ui-src/src/components/overlays/EntityDetailsOverlayV2.test.tsx
```diff
@@ -177,6 +183,46 @@ describe("<EntityDetailsOverlayV2 />", () => {
...
+
+    expect(screen.getByRole("button", { name: "Save & return" })).toBeVisible();
+  });
+
+  test("locks a closed initiative in the Work Plan", async () => {
+    mockedUseStore.mockReturnValue({
+      ...mockStateUserStore,
+      ...mockEntityStore,
+      report: { reportType: ReportType.WP } as any,
+      setAutosaveState: jest.fn(),
```
bangbay-bluetiger: `setAutosaveState` isn't being used, so don't need to add this to the mock.

services/ui-src/src/components/overlays/EntityDetailsOverlayV2.tsx
```diff
@@ -60,11 +67,12 @@ export const EntityDetailsOverlayV2 = ({
       const isClosed = isClosedInitiative(entity);
       const fields = getFields(entity);
 
-      setAutosave(!isClosed);
+      // keep autosave on in the SAR so edits to closed initiatives persist
```
bangbay-bluetiger, line 70: Good catch!

services/ui-src/src/utils/autosave/autosave.ts
```diff
@@ -137,3 +138,56 @@ export const isFieldChanged = (field: FieldInfo) => {
...
+ *
+ * To prevent that, every write is serialized through this queue. Each write
+ * is passed as a thunk and is not invoked until it reaches the front of the
+ * queue, so it always builds its payload from the result of the write before
+ * it. Explicit saves simply wait for the queue to drain.
+ */
+let writeQueue: Promise<unknown> = Promise.resolve();
+let pendingWrites = 0;
+
+const updateSavingIndicator = () => {
```
bangbay-bluetiger: I would limit scope and add a param here:
```
const updateSavingIndicator = (isAutosaving: boolean) => {
  useStore.getState?.()?.setAutosaveState(isAutosaving);
};
```

services/ui-src/src/utils/autosave/autosave.ts
```diff
@@ -137,3 +138,56 @@ export const isFieldChanged = (field: FieldInfo) => {
...
+
+/*
+ * Serializes a write behind any in-flight writes. Because `work` is a thunk
+ * invoked only when it reaches the front of the queue, it can read the newest
+ * store state when building its payload. Returns the write's own result so the
+ * caller can await it; queue progression is unaffected by its rejection.
+ */
+export const enqueueWrite = <T>(work: () => Promise<T>): Promise<T> => {
+  pendingWrites++;
+  updateSavingIndicator();
```
bangbay-bluetiger: After adding param, this can become:
```
updateSavingIndicator(true);
```

services/ui-src/src/utils/autosave/autosave.ts
```diff
@@ -137,3 +138,56 @@ export const isFieldChanged = (field: FieldInfo) => {
...
+  const run = writeQueue.then(work, work);
+  writeQueue = run
+    .catch(() => {})
+    .finally(() => {
+      pendingWrites--;
+      updateSavingIndicator();
```
bangbay-bluetiger: After adding param, this can become:
```
updateSavingIndicator(pendingWrites > 0);
```

bangbay-bluetiger (commented):
I like the new autosave methods!

services/ui-src/src/components/export/ExportedModalOverlayReportSection.tsx
ntsummers1: This is a really good call out! I'll get a ticket created to address the separation of concerns here and make sure we properly split these up

services/ui-src/src/components/export/ExportedModalOverlayReportSection.tsx
ntsummers1: Can do! Updated :D

ntsummers1 (approved):

bangbay-bluetiger (approved):

PR #1521, opened 2026-07-24
CMDCT 6351: Add link styling to "Target Populations"

amanda-forrester:
### Description
Under "Target populations" hint text, there is a link that is missing link styling. The only indicator that this a link is if a user happens to hover over the text. This is important context for usability and accessibility.

Final Fix: moved the fix into parsing.ts, where all internalLink/custom HTML elements are rendered. internalLink types now default to `className: "ds-c-link"` (CMS Design System's link class, which provides color, hover, and visited states via CSS, not just inline styles) unless an explicit className is passed in props, in which case that value will take precedence. 

This is a global fix, so it applies to every internalLink in the app consistently, not just the one field the original ticket flagged. As part of this, removed inline style={{ textDecoration: "underline" }} from other internalLink segments across the codebase (e.g. "previous page," "add FMAP Percentages"), since .ds-c-link fully handles styling on its own. Added tests as well.

~~OG Fix: Added `className: "ds-c-link"` to the `props`.~~

~~Note: There is a similar `internalLink` in the codebase...the "previous page" link in "Add initiative" that just has `style={{ textDecoration: "underline" }}`. However, it isn't nested inside hint text, so its default link styling wasn't overridden so `style={{ textDecoration: "underline" }}` was sufficient there. The "Transition Benchmarks" link lives inside hint text (`.ds-c-hint`), which applies a muted color via CSS inheritance that inline styles can't override, and inline styles also can't express :hover/:visited states. Used `className: "ds-c-link"` (the CMS Design System's link class) instead, which overrides the inherited hint color and provides real hover/visited states through actual CSS rather than inline style.~~

### Related ticket(s)

CMDCT-6351

---
### How to test
- Checkout this branch
- Open a WP in MFP
- Navigate to State or Territory Specific Initiatives and create an initiative
- Scroll down inside the initiative and make sure the words "Transition Benchmarks" under Target Population(s) has the appropriate link styling and function (see screenshot) 

Screenshot 2026-07-23 at 7 43 48 PM

### Notes

---
### Pre-review checklist

---
#### Review

#### Security


ntsummers1 (dismissed):

services/app-api/forms/routes/wp/state-or-territory-specific-initiatives/initiatives.ts
```diff
@@ -238,6 +238,7 @@ export const initiativesRoute: WPStateOrTerritorySpecificInitiativesRoute = {
                   content: "Transition Benchmarks",
                   props: {
                     to: "/wp/transition-benchmarks",
+                    className: "ds-c-link",
```
bangbay-bluetiger: Since this class will only be added to new reports and not existing ones, I think we should try adding the class globally where we generate the link element, somewhere around here:
https://github.com/Enterprise-CMCS/macpro-mdct-mfp/blob/9758fd773492133c4999d03dee9921865af7692f/services/ui-src/src/utils/other/parsing.ts#L87-L90

You could do something like this:
```
const elementProps = {
  key: type + uuid(),
  as,
  className: type === "internalLink" ? "ds-c-link" : "",
  ...props,
};
```
If className is defined in props, it'll override what is hardcoded there. But, I think when links have just been underlined without other styles, that was an oversight and we should make the links consistent.

bangbay-bluetiger (changes requested):

services/app-api/forms/routes/wp/state-or-territory-specific-initiatives/initiatives.ts
amanda-forrester: @bangbay-bluetiger yeah I like the idea of keeping it all consistent.

bangbay-bluetiger (commented):
LGTM. There's a merge conflict. I can re-review after that's addressed.

amanda-forrester:
> LGTM. There's a merge conflict. I can re-review after that's addressed.

@bangbay-bluetiger got the merge conflict resolved; thanks!

karla-vm (approved):

bangbay-bluetiger (approved):

